PR #97, opened 2019-04-05
Availability Model

msohcw:


csm_web/logistinator/models.py
```diff
@@ -1,3 +1,108 @@
...
+    """
+    Byte order is little endian, but the days are big endian.
+
+    000 ... 111
+    MON ... SUN
+
+    Within a day, it is also little endian
+
+    8AM 9AM ... 12AM
+    000 111 ... 111
```
mudit2103: Each hour corresponds to **four** bits not three, right?

csm_web/logistinator/models.py
```diff
@@ -1,3 +1,108 @@
...
+    """
+
+    # TODO Is there a cleaner way of doing this?
+    MON = 0
```
mudit2103: Use `calendar.weekday`

csm_web/logistinator/models.py
```diff
@@ -1,3 +1,108 @@
...
+    MINS_PER_INTERVAL = 60 // INTERVALS_PER_HOUR
+
+    INTERVAL = datetime.timedelta(minutes=MINS_PER_INTERVAL)
+    START_TIME = datetime.time(7, tzinfo=timezone.get_current_timezone())  # 7AM
+    END_TIME = datetime.time(23, tzinfo=timezone.get_current_timezone())  # 11PM
+
+    INTERVAL_COUNT = INTERVALS_PER_HOUR * HOURS_PER_DAY * DAYS
+
+    # By default, all slots are available
+    default_bitstring = (1 << INTERVAL_COUNT) - 1
```
mudit2103: Think it would be safer to make everything unavailable by default instead...

csm_web/logistinator/models.py
```diff
@@ -1,3 +1,108 @@
...
+
+        availabilities = tuple(
+            ((day_availability >> i) % 2) for i in range(start_index, end_index)
+        )
+        return availabilities
+
+    def get_day_availability(self, day):
+        assert 0 <= day < Availability.DAYS
+        intervals_per_day = Availability.INTERVALS_PER_HOUR * Availability.HOURS_PER_DAY
+        one_day = 1 << intervals_per_day
```
mudit2103: This bit twiddling needs documentation.

csm_web/logistinator/models.py
msohcw: This was just arbitrary but I can see why it will be confusing.

csm_web/logistinator/models.py
msohcw: ok

PR #101, opened 2019-04-07
Improved setup

Svetlitski:
Spent like an hour today getting JMs set up again after something happened with virtualenvs. I've refactored the setup script, made it more robust, and added additional functionality to it  in order to avoid having to do any setup manually in the future.

### DIFF
--- a/csm_web/scheduler/factories.py
+++ b/csm_web/scheduler/factories.py
@@ -275,7 +275,7 @@ def generate_test_data(complicate=False):
     if not settings.DEBUG:
         print("This cannot be run in production! Aborting.")
         return
-    management.call_command("flush", interactive=True)
+    management.call_command("flush", interactive=False)
     original_signals = disable_signals()
     course_names = ("CS70", "CS61A", "CS61B", "CS61C", "EE16A")
     print("Generating test data...")
--- a/csm_web/scheduler/factories.py
+++ b/csm_web/scheduler/factories.py
@@ -271,10 +271,23 @@ def reenable_signals(original_signals):
         signal.receivers = receivers
 
 
-def generate_test_data(complicate=False):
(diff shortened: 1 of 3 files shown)


csm_web/scheduler/factories.py
```diff
@@ -275,7 +275,7 @@ def generate_test_data(complicate=False):
     if not settings.DEBUG:
         print("This cannot be run in production! Aborting.")
         return
-    management.call_command("flush", interactive=True)
+    management.call_command("flush", interactive=False)
```
noloerino: I think it's still worth having interactive=True, because even though the if statement above stops you from running this in production, this check has stopped me from accidentally wiping the database on my test environment a couple times.

setup.sh
```diff
@@ -2,48 +2,79 @@
...
-[ ! -f 'package.json' ] && echo 'You are in the wrong directory!' && exit 1
-[ -z "$VIRTUAL_ENV" ] && echo 'You must activate your virtualenv first!' && exit 1
+{ [ ! -d 'csm_web' ] || [ ! -f 'package.json' ]; } && echo 'You are in the wrong directory!' 1>&2 && exit 1
+[ -z "$VIRTUAL_ENV" ] && echo 'You must activate your virtualenv first!' 1>&2 && exit 1
+
+target_version=$(sed 's/[^0-9.]//g' runtime.txt)
+installed_version=$(python3 --version | sed 's/[^0-9.]//g')
+if [ "$target_version" != "$installed_version" ]
+then
+   echo "You have python version $installed_version installed, but the expected version is $target_version.
```
noloerino: Echo this at the end instead so it doesn't get lost in the output of npm install etc.

noloerino:
@Svetlitski bump

PR #106, opened 2019-04-15
Add attendance filter

hkdragon9:


csm_web/scheduler/admin.py
```diff
@@ -318,6 +322,44 @@ def number_of_senior_mentors(self, obj):
...
+            (k, v)
+            for k, v in changelist.get_filters_params().items()
+            if k != self.parameter_name
+        )
+        yield all_choice
+
+
+class DayStartFilter(InputFilter):
+    parameter_name = "day_start"
+    title = "Day_start(YYYY-MM-DD)"
```
noloerino: Rename to "Filter by Date (YYYY-MM-DD)"

csm_web/scheduler/admin.py
```diff
@@ -318,6 +322,44 @@ def number_of_senior_mentors(self, obj):
...
+
+    def queryset(self, request, queryset):
+        if self.value() is not None:
+            filter_day_start = self.value()
+            return queryset.filter(Q(date__gte=filter_day_start))
+
+
+class DayEndFilter(InputFilter):
+    parameter_name = "week_end"
+    title = "Day_end(YYYY-MM-DD)"
```
noloerino: Rename to "Filter Before Date (YYYY-MM-DD)"

csm_web/scheduler/admin.py
```diff
@@ -318,6 +322,44 @@ def number_of_senior_mentors(self, obj):
...
+
+    def queryset(self, request, queryset):
```
noloerino: Make sure this fails nicely: right now it shows the user a 500 error if they type the date string wrong. I think the best way to handle a bad date string is:
(1) wrap the `queryset.filter` in a try/except block
(2) when the exception is triggered, catch it and tell the admin page to display a small message prompting the user to match the proper date string
If the above is too difficult to implement, I think it's better to just return an empty queryset after catching the exception than to display a scary error.

noloerino: Actually, I realized there might be an even more idiomatic way of constraining the input to YYYY-MM-DD: Django probably has some kind of date input field, that will generate the appropriate kind of HTML element (see here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date). See if you can look into that.

csm_web/scheduler/admin.py
```diff
@@ -322,12 +322,14 @@ def number_of_senior_mentors(self, obj):
...
 
 
-class InputFilter(admin.SimpleListFilter):
+class DayStartFilter(admin.SimpleListFilter):
     template = "admin/input_filter.html"
 
+    title = "start date"
+
     def lookups(self, request, model_admin):
         # Dummy, required to show the filter.
```
noloerino: Remove this comment (since it's no longer relevant)

csm_web/scheduler/admin.py
```diff
@@ -339,20 +341,33 @@ def choices(self, changelist):
...
-class DayEndFilter(InputFilter):
+class DayEndFilter(admin.SimpleListFilter):
+    template = "admin/input_filter.html"
+
     parameter_name = "week_end"
-    title = "Day_end(YYYY-MM-DD)"
+    title = "end date"
+
+    def lookups(self, request, model_admin):
+        # Dummy, required to show the filter.
```
noloerino: same as above

PR #118, opened 2019-06-18
Up Django to 2.2

noloerino:
Updating bc of a security vuln. I haven't tested this extensively (I only ran the setup script + went to a few pages on my local build), but everything seems to work, and even if it doesn't nobody cares since it's summer and we might be migrating to AWS anyway.


Svetlitski:
Sounds good, I also updated Django rest framework so that any incompatibilities are avoided

PR #126, opened 2019-08-29
Redesign models

Svetlitski:
Redesign the models underlying Scheduler to address pain points we ran into last year.
Validation has been added to several models to ensure we do not end up with nonsensical/impossible data in the database again, and test cases have been created for this validation.


csm_web/scheduler/models.py
```diff
@@ -1,25 +1,30 @@
...
     pass
 
 
-class ActivatableModel(models.Model):
-    active = models.BooleanField(default=True)
+class ValidatingModel(models.Model):
+    """
+    By default, Django models do not validate on save!
+    This abstract class fixes that insanity
+    """
```
noloerino: I thought this wasn't true, but then I checked the docs: https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.full_clean
just want to say, yikes

csm_web/scheduler/models.py
```diff
@@ -30,150 +35,118 @@ class Attendance(models.Model):
...
-    @property
-    def students(self):
-        return self.profile_set.filter(role=Profile.STUDENT)
-
-    @property
-    def mentor(self):
-        mentor_profiles = mentor_profiles = self._get_mentor_set()
-        assert mentor_profiles.count() <= 1
-        return mentor_profiles.first()
+    mentor = models.ForeignKey(Mentor, on_delete=models.SET_NULL, blank=True, null=True)
```
noloerino: do we want mentor to be a onetoone, since that's a valid assumption as long as we don't have to deal with coteaching?

also, I'm kind of wary of having a nullable field here, where use cases looking up a student's mentor's name through their section might run into problems
instead, to represent mentorless sections (if we even want to do that) we could consider creating mentor profiles that point to a "sentinel" user

csm_web/scheduler/models.py
```diff
@@ -30,150 +35,118 @@ class Attendance(models.Model):
...
+        unique_together = ("user", "section")
 
-    @property
-    def name(self):
-        return f"{self.user.first_name} {self.user.last_name}"
 
-    def __str__(self):
-        return f"{self.course.name} {self.role} - {self.name}"
+class Mentor(Profile):
+    pass
```
noloerino: we may want to also add
```python
class Coordinator(Profile):
    pass
```
to deal with coordinator permissions, although that can probably wait until we actually decide to implement it

csm_web/scheduler/models.py
```diff
@@ -30,150 +35,118 @@ class Attendance(models.Model):
...
-        return self.mentor
-
-    @property
-    def active_students(self):
-        return self.students.filter(active=True)
+    def clean(self):
+        super().clean()
+        if self.students.count() > self.capacity:
+            raise ValidationError(
+                "Number of enrolled students exceeds section capacity"
```
noloerino: there are use cases (albeit somewhat hacky ones) where we DO want to enroll over capacity
for example, to handle EOP sections, I set their max capacity to 0 and enrolled students manually: raising the capacity doesn't work, as if one of those students drops it would allow another non-EOP student to enroll through the site

instead, we should have this validation in serializer or views

csm_web/scheduler/models.py
```diff
@@ -30,150 +35,118 @@ class Attendance(models.Model):
...
-        return f"{self.user.first_name} {self.user.last_name}"
 
-    def __str__(self):
-        return f"{self.course.name} {self.role} - {self.name}"
+class Mentor(Profile):
+    pass
 
 
-class Section(models.Model):
+class Section(ValidatingModel):
```
noloerino: add a "comment" text field so we can add special notes about a section, like "EOP" or "early start" or whatever

csm_web/scheduler/models.py
```diff
@@ -183,10 +156,41 @@ class Spacetime(models.Model):
...
+    _start_time = models.TimeField()
+    _duration = models.DurationField()
+    _day_of_week = models.CharField(max_length=3, choices=DAY_OF_WEEK_CHOICES)
+
+    """
+    Unfortunately the Django models.Model class doesn't play nice with standard Python metaprogramming
+    functionality like __getattribute__, and Django doesn't allow you to shadow fields in a descendant class,
+    so we have to resort to this boilerplate @property solution in order to have Spacetime 'magically' return
+    the overriden value if there is one
+    """
```
noloerino: we may want some way to expose the original spacetime and whether or not this has an override, so on the frontend we can avoid confusing people about whether an updated time is weekly or temporary

Svetlitski: You can still access to original spacetime fields, since Python doesn't _actually_ have private methods/fields and the leading underscore is just a convention. The `has_valid_override` method seems to address your second point. Am I misunderstanding, and if so what other functionality did you want?

csm_web/scheduler/models.py
Svetlitski: Let's add it when we need it

csm_web/scheduler/models.py
Svetlitski: My rational behind making it nullable was for the following situations:
- Sometimes coords create sections they know they _will_ find people to teach and allow students to enroll in them, and then assign a mentor after students have enrolled
- A mentor gets sick or leaves CSM or has a conflict or something, so mentor may be null in the interim while coords find a replacement mentor for the section

No one in exec has displayed any interest in co-teaching, so I say we put that on the back-burner for now and leave it one-to-one.

And why can't we just solve the "looking up mentor name problem" in the serializer (i.e. `name = student.section.mentor.name if student.section.mentor else 'TBD'`)

csm_web/scheduler/models.py
Svetlitski: Interesting. Maybe we should consider adding a `public = BooleanField()` to `Section` instead of this hack? Thoughts?

csm_web/scheduler/models.py
Svetlitski: I worry about this being abused in place of building new features. EOP I understand (see above comment about `public` field on Section in place of this), but what other purposes would this serve?

csm_web/scheduler/models.py
noloerino: my issue is there might be some other use case where we need to look up a field on a null mentor, and having to remember to handle the null in every case is rather annoying

granted, it probably is more semantically clear to have a section w/o an assigned teacher have a null mentor, so this is fine as is i guess

csm_web/scheduler/models.py
noloerino: i think it's better to have some extra flexibility in this regard, since there may be more potential use cases (i think there was one other situation where it would have been helpful)

csm_web/scheduler/models.py
noloerino: i think it's better to keep the system as simple as possible, and if we don't need to add additional fields to serve a purpose we shouldn't - my idea is that the "comment" field shouldn't be used to control any logic, just as a way of displaying some extra information

i can see how it could be potentially abused; that said, i think django is easy enough to modify that if we see it getting out of control, we would start feature work pretty quickly - in general, i don't think we should spend more time engineering new solutions than is really necessary

csm_web/scheduler/models.py
noloerino: hm that's true, although i was thinking of having a way to access the original field without breaking that convention

`has_valid_override` is good, but i don't see it in the diff: can you give me a line number?

noloerino: oh just kidding it just showed up

csm_web/scheduler/models.py
Svetlitski: Okay, will revert

csm_web/scheduler/models.py
Svetlitski: Will add

PR #128, opened 2019-08-30
Remove logistinator and update dependencies

Svetlitski:
Update JS and python dependencies, bump python version to 3.7.3
Remove logistinator code


noloerino:
lgtm, if there's any leftover issues wrt not deleting stuff properly we can resolve it later anyway

PR #129, opened 2019-08-30
Replace black with autopep8

Svetlitski:
`black` produces code that is miserable to read, so we're replacing it with `autopep8`.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -59,7 +59,6 @@ python csm_web/manage.py createtestdata --yes
 
 # TODO put precommit hooks in git?
 echo "Installing pre-commit hook..."
-pip3 install black # pre-commit hook requires this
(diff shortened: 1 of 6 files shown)


setup.sh
```diff
@@ -59,7 +59,6 @@ python csm_web/manage.py createtestdata --yes
 
 # TODO put precommit hooks in git?
 echo "Installing pre-commit hook..."
-pip3 install black # pre-commit hook requires this
 ln -s -f ../../.pre-commit.sh .git/hooks/pre-commit
```
noloerino: do we want pip3 install autopep8 in the setup script?

Svetlitski: I'd rather have people install it globally so that if anyone ever updates requirements besides us we don't include `autopep8` (which is a dev dependency) in the requirements.txt

noloerino: sure, sounds reasonable

PR #130, opened 2019-09-05
Serializers 2.0

Svetlitski:
New serializers for the refactored models.
Other unrelated changes that are bundled in this PR:
- Fix Mentor `__str__` method
- Add `active` to `Student`, add `permitted_absences` to `Course`
- Add `django_extensions` for dev quality of life


csm_web/csm_web/settings.py
```diff
@@ -36,10 +36,9 @@
 
 ALLOWED_HOSTS = []
 
-# Change this when Matthew graduates I guess
 ADMINS = [
     ("Jonathan Shi", "jhshi@berkeley.edu"),
-    ("Matthew Soh", "matthewsoh@berkeley.edu"),
+    ("Kevin Svetlitski", "kevin_svetlitski@berkeley.edu"),
```
noloerino: lmao nice catch

csm_web/csm_web/settings.py
```diff
@@ -59,6 +58,13 @@
     "django_filters",
 ]
 
+if DJANGO_ENV == DEVELOPMENT:
+    INSTALLED_APPS.append("django_extensions")
```
noloerino: do you think we might want this in production too? i ended up spending a fair bit of time in shell in production

Svetlitski: Good point

PR #131, opened 2019-09-05
Update admin page

noloerino:
Updated the Django admin page to reflect our changes to the schema. This mostly involved renaming fields and splitting profile into mentor/student (which really helped).

This commit only adds most basic functionality, and I plan to add on fancier stuff (autocomplete, more robust search filters, single forms for student enrollments/section creation and the like), but I'd like to first get all this stuff merged.

One thing I'd like feedback on: I created a `Coordinator` model with an FK to `Course` so we could make it so coordinators can only view/modify for their own course. Strictly speaking, we don't need to set permissions (since if we have a malicious coord user we have bigger problems than scheduler), but making it so coords only see their own course's resources by default makes it easier to avoid confusion. This also allows us to allow other users to make changes without superusering them, by giving them coordinator profiles.


csm_web/scheduler/models.py
```diff
@@ -92,23 +92,51 @@ class Meta:
...
+        help_text="The recurring time and location of a section. This can be temporarily overriden "
+        "by the mentor, in which case the admin page will display the overriding times."
+    )
     capacity = models.PositiveSmallIntegerField()
     mentor = models.OneToOneField(Mentor, on_delete=models.SET_NULL, blank=True, null=True)
-    description = models.CharField(max_length=100, blank=True)
+    description = models.CharField(
+        max_length=100,
+        blank=True,
+        help_text="A brief note to add some extra information about the section, e.g. \"EOP\" or "
```
Svetlitski: Use single quotes here instead of escaping
```python
help_text='A brief note to add some extra information about the section, e.g. "EOP" or "Early start".'
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+
+    drop_students.short_description = (
+        "Drop student(s) from a section. This action is reversible, but it is still strongly "
+        "recommended that you double check the name/email/section/course first."
+    )
+
+    def undrop_students(self, request, queryset):
+        queryset.update(active=True)
+
+    undrop_students.short_description = "Re-enroll student(s) in the section they were previously dropped from."
```
Svetlitski:
```suggestion
    reenroll_students.short_description = "Re-enroll student(s) in the section they were previously dropped from."
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
         return request.user.is_superuser
 
 
+@admin.register(Student)
+class StudentAdmin(CoordAdmin):
```
Svetlitski: Because you have `get_email`, `get_course`, `get_mentor`, and `section` in `list_display`, the admin will have to make a SQL query for each of the related objects for each Student, which is **terrible** for performance, even on a small dataset you can overwhelm the connection pool like this.

To remedy this, preload the relations for each object in the query set, which drastically reduces the number of queries made:
```python
def get_queryset(self, request):
    return super().get_queryset(request).select_related('user', 'section__course', 'section__mentor')
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+
+    get_user.short_description = "User"
+
+    # TODO handle nullable section
+    def get_course(self, obj):
+        return get_admin_link_for(obj.section.course, "admin:scheduler_course_change")
+
+    get_course.short_description = "Course"
+
+    def get_mentor(self, obj):
```
Svetlitski: Missing `short_description` for `get_mentor`. Is this intentional?

csm_web/scheduler/admin.py
```diff
@@ -8,28 +8,67 @@
...
     Spacetime,
     Override,
 )
 
+admin.site.register(Coordinator)
+
+# Helper methods
+
+
+def get_admin_link_for(obj_property, reverse_path):
```
Svetlitski:
```suggestion
def get_admin_link_for(obj_property, reverse_path, link_contents = None):
```

csm_web/scheduler/admin.py
```diff
@@ -8,28 +8,67 @@
...
+    rather than "user".
+    Uses the provided str of obj_property.
+    """
+    admin_url = reverse(
+        reverse_path, args=(obj_property.id,)
+    )
+    return format_html(
+        '<a style="display: block" href="{}">{}</a>',
+        admin_url,
+        obj_property
```
Svetlitski:
```suggestion
        link_contents or obj_property
```

csm_web/scheduler/admin.py
Svetlitski: Modified to generalize for use by attendances in `Student` admin, see below

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+
+    get_course.short_description = "Course"
+
+    def get_mentor(self, obj):
+        return get_admin_link_for(obj.section.mentor, "admin:scheduler_mentor_change")
+
+    def get_attendances(self, obj):
+        attendance_links = []
+        for attendance in obj.attendance_set.all():
+            admin_url = reverse(
```
Svetlitski: Let's reuse `get_admin_link_for` instead of repeating ourselves.
Double check me, but I believe this would be:
```python
def get_attendances(self, obj):
    attendance_links = sorted(get_admin_link_for(attendance, 'admin:scheduler_attendance_change', f"Date {attendance.date}: {attendance.presence or '--'}") for attendance in obj.attendance_set.all())
    return format_html("".join(attendance_links))
```
Note the lack of list brackets inside `attendance_links` is deliberate, that's a generator comprehension if you haven't seen that before.

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+    actions = ("deactivate_profiles", "activate_profiles")
+    # autocomplete_fields = ("section", "user")
+
+    def has_delete_permission(self, request, obj=None):
+        return (
+            False
+        )  # delete sections + mentor by deleting sections, drop by deactivating
+
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
```
Svetlitski: Change `qs` to `queryset` for clarity and consistency, preload reload relations (explained above for student case)
```suggestion
        queryset = super().get_queryset(request).select_related('user', 'section__course')
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+        if request.user.is_superuser:
+            return qs
+        return qs.filter(section__course__in=get_visible_courses(request.user))
+
+    def get_fields(self, request, obj):
+        fields = ["name", "get_email", "get_course", "get_section", "get_students"]
+        fields.insert(4, "user" if request.user.is_superuser else "get_user")
```
Svetlitski:
```suggestion
        fields.append("user" if request.user.is_superuser else "get_user")
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+        return tuple(fields)
```
Svetlitski:
```suggestion
        return fields
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+
+    def get_readonly_fields(self, request, obj):
+        fields = ["get_course", "get_email", "get_section", "name", "get_students"]
+        if not request.user.is_superuser:
+            fields.insert(0, "get_user")
```
Svetlitski:
```suggestion
            fields.append("get_user")
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+        return tuple(fields)
```
Svetlitski:
```suggestion
        return fields
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+
+    get_user.short_description = "User"
+
+    # TODO handle nullable section
+    def get_section(self, obj):
+        return get_admin_link_for(obj.section, "admin:scheduler_section_change")
+
+    get_section.short_description = "Section"
+
+    def get_course(self, obj):
```
Svetlitski: Reuse `get_admin_link_for`. See above example with attendances.

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+        )
+        return format_html(
+            '<a style="display: block" href="{}">{}</a>',
+            admin_url,
+            course
+        )
+
+    get_course.short_description = "Course"
+
+    def get_students(self, obj):
```
Svetlitski:
```python
def get_students(self, obj):
    return format_html("".join(sorted(get_admin_link_for(student, "admin:scheduler_student_change") for student in obj.section.students.all())))
```

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
+        for student in obj.section.students.all():
+            student_links.append(get_admin_link_for(student, "admin:scheduler_student_change"))
+        student_links.sort()
+        return format_html("".join(student_links))
+
+    get_students.short_description = "Students"
+
+
 @admin.register(Section)
 class SectionAdmin(CoordAdmin):
```
Svetlitski: See above comments about changing `get_queryset`

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
-        admin_url = reverse("admin:scheduler_profile_change", args=(obj.mentor.id,))
-        return format_html('<a href="{}">{}</a>', admin_url, obj.mentor.name)
+    def get_mentor_name(self, obj):
+        return obj.mentor.name
 
-    get_mentor_display.short_description = "mentor"
+    get_mentor_name.short_description = "Name"
 
-    def students(self, obj):
+    def get_students(self, obj):
```
Svetlitski: Fix similar to other methods I commented on above

csm_web/scheduler/admin.py
```diff
@@ -49,273 +88,311 @@ def has_delete_permission(self, request, obj=None):
...
     def has_delete_permission(self, request, obj=None):
         return False  # will break 1to1 invariants if deleted
 
-    def get_queryset(self, request):
-        return Spacetime.objects.all()
-
 
 @admin.register(Course)
-class CourseAdmin(admin.ModelAdmin):
+class CourseAdmin(CoordAdmin):
```
Svetlitski: Again with fixing the `get_queryset` to prefetch

csm_web/scheduler/admin.py
```diff
@@ -378,106 +455,90 @@ def queryset(self, request, queryset):
...
         "presence",
     )
 
-    list_filter = ("presence", "attendee__course", DayStartFilter, DayEndFilter)
-    search_fields = ("attendee__user__first_name", "attendee__user__last_name")
+    list_filter = ("presence", "student__section__course", DayStartFilter, DayEndFilter)
+    search_fields = ("student__user__first_name", "student__user__last_name")
     ordering = ("-date",)
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
```
Svetlitski:
```python
def formfield_for_foreignkey(self, db_field, request, **kwargs):
    if not request.user.is_superuser and db_field.name == "student":
        kwargs["queryset"] = Student.objects.filter(student__section__course__in=get_visible_courses(request.user))
    return super().formfield_for_foreignkey(db_field, request, **kwargs)
```

csm_web/scheduler/admin.py
```diff
@@ -378,106 +455,90 @@ def queryset(self, request, queryset):
...
+        if not request.user.is_superuser:
+            courses = get_visible_courses(request.user)
+            if db_field.name == "student":
+                kwargs["queryset"] = Student.objects.filter(student__section__course__in=courses)
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
+
+    def get_queryset(self, request):
```
Svetlitski: `qs`->`queryset`

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+class StudentAdmin(CoordAdmin):
+    list_filter = ("active",)
+    list_display = ("name", "get_email", "get_course", "get_mentor", "section")
+    search_fields = ("user__email", "user__first_name", "user__last_name")
+    actions = ("drop_students", "undrop_students")
+    # autocomplete_fields = ("section", "user")
+
+    def get_fields(self, request, obj):
+        fields = ["name", "get_email", "get_course", "section", "get_attendances"]
+        fields.insert(4, "user" if request.user.is_superuser else "get_user")
```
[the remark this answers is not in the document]
noloerino: I'd prefer to keep the insert since this determines the order in which fields appear

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+        return tuple(fields)
+
+    def get_readonly_fields(self, request, obj):
+        fields = ["get_course", "get_email", "name", "get_attendances"]
+        if not request.user.is_superuser:
+            fields.insert(0, "get_user")
+        return tuple(fields)
+
+    def has_delete_permission(self, request, obj=None):
```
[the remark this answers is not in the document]
noloerino: The precommit hook definitely ran (I saw terminal output), not sure how this slipped through

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        if request.user.is_superuser:
+            return qs
+        return qs.filter(section__course__in=get_visible_courses(request.user))
+
+    def drop_students(self, request, queryset):
+        queryset.update(active=False)
+
+    drop_students.short_description = (
```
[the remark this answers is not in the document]
noloerino: oh yeah it's quite nice

csm_web/scheduler/admin.py
```diff
@@ -49,273 +96,311 @@ def has_delete_permission(self, request, obj=None):
...
+        "Drop student(s) from a section. This action is reversible, but it is still strongly "
+        "recommended that you double check the name/email/section/course first."
+    )
+
+    def undrop_students(self, request, queryset):
```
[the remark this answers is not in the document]
noloerino: I think it's important to make clear that this action is the exact inverse of "drop" - "reenroll" might carry with it the implication that it would let you enroll a student in a different section

csm_web/scheduler/admin.py
noloerino: I initially thought that this field wasn't being displayed since I was using it for some search thing but that seems not to be the case, oops

csm_web/scheduler/admin.py
noloerino: I don't think it's possible here, since all the references here are reverse foreign keys

csm_web/scheduler/admin.py
```diff
@@ -8,28 +8,75 @@
...
+        return Course.objects.all()
+    else:
+        return Coordinator.objects.filter(user=user).values_list("course")
+
+
+def is_user_admin(user):
+    return user.is_superuser or user.is_staff or get_obj_or_none(Coordinator, user=user) is not None
+
 
 class CoordAdmin(admin.ModelAdmin):
```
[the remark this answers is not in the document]
noloerino: this disgusts me but sure

csm_web/scheduler/admin.py
Svetlitski: Have to use `prefetch_related`, which is the sibling to `select_related`: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#prefetch-related

noloerino: Neat. Also added this to MentorAdmin and SectionAdmin for student set lookup.

Svetlitski:
Looks good to me!

PR #132, opened 2019-09-07
Views 2.0

Svetlitski:


csm_web/scheduler/serializers.py
```diff
@@ -27,7 +27,10 @@ class Meta:
 class AttendanceSerializer(serializers.ModelSerializer):
     class Meta:
         model = Attendance
-        fields = ("id", "presence", "week_start")
+        fields = ("id", "presence", "week_start", "date", "student")
+        read_only_fields = ("week_start",)
+        extra_kwargs = {'student': {'write_only': True},
```
Svetlitski: Isn't is annoying that DRF doesn't have a `write_only_fields` for its `Meta` classes?

noloerino: yeah that's pretty annoying

csm_web/scheduler/views.py
```diff
@@ -1,267 +1,150 @@
...
+            section = Section.objects.select_for_update().get(pk=section.pk)
+            if section.current_student_count >= section.capacity:
+                raise PermissionDenied("There is no space available in this section", status.HTTP_423_LOCKED)
+            try:  # Student dropped a section in this course and is now enrolling in a different one
+                student = Student.objects.get(active=False, section__course=section.course, user=request.user)
+                student.section = section
+                student.active = True
+                student.save()
+                return Response(status=status.HTTP_204_NO_CONTENT)
+            except Student.DoesNotExist:  # Student is enrolling in this course for the first time
```
noloerino: Wouldn't this case also cover a student attempting to change sections?

csm_web/scheduler/views.py
```diff
@@ -1,267 +1,150 @@
...
-                "This profile ({}) has been deactivated".format(profile)
-            )
-        else:
+    def get_queryset(self):
+        return Spacetime.objects.filter(section__mentor__user=self.request.user)
+
+    @action(detail=True, methods=['put'])
+    def override(self, request, pk=None):
+        spacetime = get_object_or_error(self.get_queryset(), pk=pk)
+        if hasattr(spacetime, "override"):  # update
```
noloerino: I feel like there should be some kind of abstraction in models for this: perhaps add `has_any_override` to Spacetime?

csm_web/scheduler/views.py
Svetlitski: I don't understand. Could you elaborate?

csm_web/scheduler/views.py
Svetlitski: We do have `has_valid_override`, but here I'm specifically testing to see that an override exists, even if it's expired.

noloerino: Yeah, but I don't like breaking the abstraction barrier with hasattr here.

csm_web/scheduler/views.py
noloerino: If a student already in a section in this course attempts to enroll in a section, then the `Student.objects.get` query would error, right? So we'd end up in this branch, where we attempt to create a new student?

Svetlitski: No, the `Student.objects.get` query only errors if no matching object is found. No matching object found => the current user does not already have a student profile in this course. Which brings us to the except branch where we then create a student profile for them since they don't already have one.

noloerino: The `active=False` in the query should cause such an error afaik

PR #133, opened 2019-09-07
Logging

Svetlitski:
Add basic logging for the most critical events, remove django-eventlog and corresponding signals cruft.


csm_web/scheduler/views.py
```diff
@@ -10,6 +10,9 @@
 from rest_framework import mixins
 from .serializers import CourseSerializer, SectionSerializer, StudentSerializer, AttendanceSerializer, MentorSerializer, OverrideSerializer
 from django.core.exceptions import ObjectDoesNotExist
+import logging
+
+logger = logging.getLogger(__name__)
```
noloerino: Something like "scheduler.views" would be better instead of __name__, especially if we have more apps in the future where we log from views

csm_web/scheduler/views.py
```diff
@@ -73,21 +76,31 @@ def students(self, request, pk=None):
             the section directly, any student trying to enroll must first acquire a lock on the
             desired section. This allows us to assume that current_student_count is correct.
             """
+            section = get_object_or_error(Section.objects, pk=pk)
             section = Section.objects.select_for_update().get(pk=section.pk)
             if not request.user.can_enroll_in_course(section.course):
+                logger.warn(
+                    f"<Enrollment:Failure> User {request.user.pk} was unable to enroll in Section {section.pk} because they are already involved in this course")
```
noloerino: Would be useful to display the __str__ as well in case the section mutates

noloerino: (Also makes much more readable, esp if we're searching the logs for users - saves us the time of going to django shell to check)

noloerino: Also applies to all the below logs, maybe have a helper function to output str + pk of a model

csm_web/scheduler/views.py
Svetlitski: This is dynamically populated with the value `scheduler.views`. `__name__` refers to the name of the current module in python.

noloerino: Ah, I thought it would be the name of the file

(also, didn't realize that __ meant bold in markdown)

csm_web/scheduler/views.py
Svetlitski: Agree on printing user info (email address since that's unique), but don't understand why we need to print anything except Section pk. Other attributes can mutate, but the pk never will.

noloerino: My point is that if other attributes mutate, this log is no longer useful, since we care about the state of the system at the time of the log event

PR #134, opened 2019-09-11
Frontend 2.0

Svetlitski:
Sometimes things have to get worse before they can get better, like the UI for the entire application.


noloerino:
i'm fine with merging, but i think there's 2 semi-critical things we should add:
(1) add ability to see multiple sections (perhaps hide attendance for now, and just display all profiles sequentially)
(2) courses page stuff needs to be spaced a lot better:
Screen Shot 2019-09-10 at 18 07 04
maybe just add padding/margins to everything?
there also needs to be some indication of the current selected course
(3) a drop button/hyperlink should be exposed to allow people to drop/swap sections

noloerino:
(oops i had cached css - the rest of the comments still apply though)

PR #136, opened 2019-09-14
Admin fixes

noloerino:
I gave up on making spacetime creation in sections work for now, but this commit adds a bunch of qol stuff like being able to filter students on course, and making a few fields autocompletes rather than dropdowns.

### DIFF
--- a/csm_web/scheduler/admin.py
+++ b/csm_web/scheduler/admin.py
@@ -81,11 +81,11 @@ class UserAdmin(CoordAdmin):
 
 @admin.register(Student)
 class StudentAdmin(CoordAdmin):
-    list_filter = ("active",)
+    list_filter = ("active", "section__course")
     list_display = ("name", "get_email", "get_course", "get_mentor", "section")
     search_fields = ("user__email", "user__first_name", "user__last_name")
     actions = ("drop_students", "undrop_students")
-    # autocomplete_fields = ("section", "user")
+    autocomplete_fields = ("section", "user")
 
     def get_fields(self, request, obj):
         fields = ["name", "get_email", "get_course", "section", "get_attendances", "active"]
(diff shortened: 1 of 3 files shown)


csm_web/scheduler/models.py
```diff
@@ -76,6 +78,10 @@ def clean(self):
         if self.valid_until < enrollment_end:
             raise ValidationError("valid_until must be after enrollment_end")
 
+    def is_open(self):
+        now = timezone.now()
+        return now > self.enrollment_start and now < self.enrollment_end
```
Svetlitski:
```suggestion
        return self.enrollment_start < now < self.enrollment_end
```

csm_web/scheduler/admin.py
```diff
@@ -292,21 +302,29 @@ def get_queryset(self, request):
...
-            <a class="related-widget-wrapper-link change-related"change-related" \
-            id="change_id_spacetime" \
-            href="{}?_to_field=id&_popup=1">{}</a></div>',
-            admin_url,
-            str(obj.spacetime),
-        )
+        if obj.id is not None:
+            admin_url = reverse("admin:scheduler_spacetime_change", args=(obj.spacetime.id,))
+            return format_html(
+                '<div class="related-widget-wrapper"> \
```
Svetlitski: Change to triple-quoted strings instead of all these ugly linebreaks

csm_web/scheduler/admin.py
```diff
@@ -81,11 +81,11 @@ def has_delete_permission(self, request, obj=None):
 
 @admin.register(Student)
 class StudentAdmin(CoordAdmin):
```
Svetlitski: Not important for this PR, but in the future I'd like to pull permissioning out into a separate mixin class instead of having everything inherit from CoordAdmin, which is somewhat semantically confusing

Svetlitski:
I tried my hand at implementing the inline for Spacetime on Section, but even trying to use the `django-reverse-admin` package I *still* couldn't get it to work, so I guess we'll just leave this alone for now.

csm_web/scheduler/models.py
noloerino: ugh fine

PR #138, opened 2019-09-15
Make spacetime on section admin editable

noloerino:
Creates a custom admin form so we can actually create sections with new spacetimes.
I couldn't find an easy way to embed students as a URL (short of creating a new widget), but I think it shouldn't be too difficult to find the student through admin as a workaround.

### DIFF
--- a/csm_web/scheduler/admin.py
+++ b/csm_web/scheduler/admin.py
@@ -5,6 +5,7 @@ from django.utils.html import format_html
 from django.urls import reverse
 from django.db import transaction
 from django.contrib import messages
+from django import forms
 
 from scheduler.models import (
     User,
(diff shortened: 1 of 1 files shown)


csm_web/scheduler/admin.py
```diff
@@ -235,13 +227,100 @@ def get_students(self, obj):
...
+    mentor_profile_id = forms.CharField(required=False, disabled=True)
+    spacetime = forms.ModelChoiceField(required=False, queryset=Spacetime.objects.all(), widget=forms.HiddenInput())
+    location = forms.CharField(max_length=100)
+    start_time = forms.TimeField(
+        widget=forms.TimeInput(attrs={'type': 'time'}),
+        help_text="Enter a value of the form 'hh:mm', e.g. '13:00' for a 1PM section."
+    )
+    duration = forms.DurationField(
+        widget=forms.TimeInput(attrs={'type': 'time'}),
+        help_text="Enter a value of the form 'hh:mm', e.g. '01:30' for a 1.5 hour section."
```
Svetlitski: This help text is no longer applicable now that the widget has been changed

csm_web/scheduler/admin.py
```diff
@@ -235,13 +227,100 @@ def get_students(self, obj):
...
+    mentor_email = forms.CharField(required=False, disabled=True)
+    mentor_profile_id = forms.CharField(required=False, disabled=True)
+    spacetime = forms.ModelChoiceField(required=False, queryset=Spacetime.objects.all(), widget=forms.HiddenInput())
+    location = forms.CharField(max_length=100)
+    start_time = forms.TimeField(
+        widget=forms.TimeInput(attrs={'type': 'time'}),
+        help_text="Enter a value of the form 'hh:mm', e.g. '13:00' for a 1PM section."
+    )
+    duration = forms.DurationField(
+        widget=forms.TimeInput(attrs={'type': 'time'}),
```
Svetlitski: Duration widget should **not** be set to time, this makes it display with AM/PM, which makes no sense for a duration

csm_web/scheduler/admin.py
```diff
@@ -235,13 +227,100 @@ def get_students(self, obj):
...
+                fields[field].widget = forms.HiddenInput()
+
+    mentor_name = forms.CharField(required=False, disabled=True)
+    mentor_email = forms.CharField(required=False, disabled=True)
+    mentor_profile_id = forms.CharField(required=False, disabled=True)
+    spacetime = forms.ModelChoiceField(required=False, queryset=Spacetime.objects.all(), widget=forms.HiddenInput())
+    location = forms.CharField(max_length=100)
+    start_time = forms.TimeField(
+        widget=forms.TimeInput(attrs={'type': 'time'}),
+        help_text="Enter a value of the form 'hh:mm', e.g. '13:00' for a 1PM section."
```
Svetlitski: This help text is no longer relevant (and actually misleading) now that we're using the HTML5 time input

PR #141, opened 2019-09-25
Spacetime location normalization

Svetlitski:


csm_web/scheduler/models.py
```diff
@@ -250,6 +252,10 @@ def __str__(self):
         num_minutes = int(self.duration.total_seconds() // 60)
         return f"{self.location} {self.day_of_week} {formatted_time} for {num_minutes} min"
 
+    def save(self, *args, **kwargs):
+        self._location = re.sub(self.SPACE_REDUCE_REGEX, ' ', self._location).strip().title()
```
noloerino: a concern with this kind of normalization is that sometimes you want a word to be all caps (e.g. I initially abbreviated Sutardja Dai as `SDH` which gets normalized to `Sdh`), so it might look kind of weird - i think removing whitespace is fine, but case-sensitivity should be handled by the person validating the data

Svetlitski: That makes sense. My original motivation for this PR was making it possible to query for duplicate/overlapping bookings, and I suppose that can still be accomplished with a case insensitive query.

PR #147, opened 2019-12-27
Upgrade to Django 3

noloerino:
Upgrades version for Django and a few Django-related packages. I held off on throwing in Poetry for now, as there doesn't seem to be a clean way to migrate from a requirements.txt file, and I didn't want to figure out which dependencies were actual dependencies and which were secondary.


csm_web/scheduler/factories.py
```diff
@@ -45,7 +45,7 @@ class Meta:
 
 
 BUILDINGS = ("Cory", "Soda", "Kresge", "Moffitt")
-DAY_OF_WEEK_DB_CHOICES = [db_value for db_value, display_name in Spacetime.DAY_OF_WEEK_CHOICES]
+DAY_OF_WEEK_DB_CHOICES = [db_value for db_value, display_name in Spacetime.DayOfWeek.choices]
```
Svetlitski: We don't need this comprehension anymore, `DAY_OF_WEEK_DB_CHOICES` can be replaced with `Spacetime.DayOfWeek.values` everywhere it is used.

https://docs.djangoproject.com/en/3.0/ref/models/fields/#field-choices-enum-types

csm_web/scheduler/factories.py
```diff
@@ -91,7 +91,7 @@ class Meta:
     mentor = factory.SubFactory(MentorFactory)
 
 
-PRESENCE_DB_VALUES = [db_value for db_value, display_name in Attendance.PRESENCE_CHOICES]
+PRESENCE_DB_VALUES = [db_value for db_value, display_name in Attendance.Presence.choices]
```
Svetlitski: See above comment, modify accordingly

csm_web/scheduler/factories.py
```diff
@@ -102,7 +102,7 @@ class Meta:
     student = factory.SubFactory(StudentFactory)
 
 
-DAYS = [db_value for db_value, display_name in Spacetime.DAY_OF_WEEK_CHOICES]
+DAYS = [db_value for db_value, display_name in Spacetime.DayOfWeek.choices]
```
Svetlitski: Same as above

csm_web/scheduler/factories.py
```diff
@@ -121,7 +121,7 @@ def date(obj):
     spacetime = factory.SubFactory(SpacetimeFactory)
 
 
-WEEKDAY_MAP = {number: pair[0] for number, pair in enumerate(Spacetime.DAY_OF_WEEK_CHOICES)}
+WEEKDAY_MAP = {number: pair[0] for number, pair in enumerate(Spacetime.DayOfWeek.choices)}
```
Svetlitski: Similar to above

csm_web/scheduler/models.py
```diff
@@ -173,29 +170,22 @@ class Meta:
...
-        (MONDAY, "Monday"),
-        (TUESDAY, "Tuesday"),
-        (WEDNESDAY, "Wednesday"),
-        (THURSDAY, "Thursday"),
-        (FRIDAY, "Friday"),
-        (SATURDAY, "Saturday"),
-        (SUNDAY, "Sunday"),
-    )
-    DAY_INDEX = tuple(day for day, _ in DAY_OF_WEEK_CHOICES)
+    class DayOfWeek(models.TextChoices):
```
Svetlitski: Specifying the labels here is unnecessary, if omitted Django will assume it should be the capitalized version of the enum name. So `MONDAY = "Mon"` would suffice, for example

PR #148, opened 2019-12-28
Redesign course enrollment

Svetlitski:
Complete implementation of the redesigned section enrollment flow. Note that sections you are enrolled in are not yet shown on the home page, that set of features will be tackled in a separate PR.


csm_web/frontend/src/components/Course.js
```diff
@@ -0,0 +1,180 @@
...
+
+  modalContents() {
+    const iconWidth = "10em";
+    const iconHeight = "10em";
+    if (this.state.enrollmentSuccessful) {
+      return (
+        <React.Fragment>
+          <CheckCircle color="green" height={iconHeight} width={iconWidth} />
+          <h3>Successfully enrolled</h3>
+          <button className="modal-btn" modalClose>
```
noloerino: From the console: "React does not recognize the `modalClose` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `modalclose` instead. If you accidentally passed it from a parent component, remove it from the DOM element."

After messing around a bit, it seems like `modalclose` or `modalclose="true"` both still incur a warning despite the console's advisories; `modalclose={1}` worked though.

csm_web/frontend/src/components/Course.js
```diff
@@ -0,0 +1,180 @@
...
+          </button>
+        </React.Fragment>
+      );
+    }
+    return (
+      <React.Fragment>
+        <XCircle color="red" height={iconHeight} width={iconWidth} />
+        <h3>Enrollment failed</h3>
+        <h4>{this.state.errorMessage}</h4>
+        <button className="modal-btn" modalClose>
```
noloerino: see above

csm_web/frontend/src/components/Modal.js
```diff
@@ -0,0 +1,31 @@
...
+  if (!(children instanceof Array)) {
+    children = [children];
+  }
+  return (
+    <React.Fragment>
+      <div className="modal-overlay" onClick={closeModal} />
+      <div className="modal">
+        <div className="modal-contents">
+          {children.map((child, i) =>
+            Object.prototype.hasOwnProperty.call(child.props, "modalClose") ? (
```
noloerino: see above about `modalclose`

csm_web/frontend/src/components/Course.js
```diff
@@ -0,0 +1,180 @@
...
+              onChange={({ target: { checked } }) => this.setState({ showUnavailable: checked })}
+            />
+            Show unavailable
+          </label>
+        </div>
+        <div id="course-section-list">
+          {sections[currDay]
+            .filter(({ numStudentsEnrolled, capacity }) => this.state.showUnavailable || numStudentsEnrolled < capacity)
+            .map(section => (
+              <SectionCard key={section.id} {...section} />
```
noloerino: As a quality of life thing, consider displaying a card that says "No available sections" or similar to make sure that the day of week selector doesn't change place if no sections are available.

csm_web/scheduler/views.py
```diff
@@ -73,10 +73,10 @@ def get_queryset(self):
     @action(detail=True)
     def sections(self, request, pk=None):
         course = get_object_or_error(self.get_queryset(), pk=pk)
-        if request.query_params.get('grouped'):
-            return Response({day: SectionSerializer(sorted(group, key=lambda section: section.spacetime._start_time), many=True).data for day, group in groupby(
-                course.section_set.all().order_by('spacetime___day_of_week'), lambda section: section.spacetime.get__day_of_week_display())})
-        return Response(SectionSerializer(course.section_set.all(), many=True).data)
+        #TODO: Clean this up
+        return Response(dict(sorted(((day, SectionSerializer(group, many=True).data) for day, group in groupby(
```
noloerino: `dict(sorted(...))` kind of rubs me the wrong way - could you use an `OrderedDict` instead?

csm_web/frontend/src/components/Course.js
Svetlitski: Hmm. So the thing is we _don't_ want `modalClose` to actually appear on the DOM element, things are working as intended. `modalClose` is just used by the `Modal` component to indicate that this element should close the modal when it is clicked (in addition to performing its own `onClick` event if there is one). We could do something hacky like add a special class name to the element instead of using the custom `modalClose` prop, but then we definitely leak this into the DOM, all for the sake of suppressing a warning. Not sure what the best compromise is here.

csm_web/scheduler/views.py
Svetlitski: Had the same instinct at first, but `OrderedDict` only preserves insertion order and provides a mechanism to move a key to the end of the ordering. `OrderedDict` provides no mechanism for sorting keys, but we want the keys to be sorted in weekday order for ease of consumption by the frontend. Since Python 3.7 normal dicts are guaranteed to maintain keys in insertion order, but the only way to _get_ the insertion order we want is by sorting.

csm_web/frontend/src/components/Course.js
noloerino: Is it a problem that it gets leaked onto the DOM? I was under the impression that this is the kind of thing that class names should be used for.

Svetlitski: It's not a problem per-say, it just strikes me as a little odd to have a CSS class on an element that doesn't actually apply any styling, but is used purely for JS logic.

Here's an alternative implementation I just thought of. Inside Modal.js we define
```javascript
export function ModalCloser({children}) {
    return <React.Fragment>{children}</React.Fragment>;
}
```

And you just wrap anything that should close the modal in this element instead of using the `modalClose` prop. Then inside the `Modal` component you have `child.type == ModalCloser` instead of `Object.prototype.hasOwnProperty.call(child.props, "modalClose")`

noloerino: sure, that makes sense

Svetlitski: Wait actually that has issues, it means you can't nest anything that will close the Modal inside anything else.

Svetlitski: Resolved using that idea with React context

csm_web/frontend/src/components/Course.js
Svetlitski: Added, good idea

PR #150, opened 2020-01-06
Redesign home page

Svetlitski:
The home page now displays a card for each course a user is associated with, each of which links to the appropriate section details (detail views not yet implemented, will be in next PR). I've modified the `createtestdata` management command to produce a single user `demo_user` which conveniently has a set of profiles that expose all student and mentor functionality, and expose all functionality on the home page. This is strictly better than what the command did before.


csm_web/frontend/src/components/App.js
```diff
@@ -1,62 +1,20 @@
...
-                        .filter(profile => profile.section == routeProps.match.params.id)[0].id
-                    }
-                    isMentor={mentorProfiles
-                      .map(profile => profile.section)
-                      .includes(Number(routeProps.match.params.id))}
-                    {...routeProps}
-                  />
-                )}
-              />
+              <Route path="/sections/:id" />
```
Svetlitski: This deliberately hasn't been hooked up to any component yet, as the corresponding view hasn't been implemented

csm_web/scheduler/models.py
```diff
@@ -95,7 +96,7 @@ class Meta:
 class Student(Profile):
     """
     Represents a given "instance" of a student. Every section in which a student enrolls should
-    have a new Mentor profile.
```
Svetlitski: Just fixing some random typos I noticed, not related to the functionality added in this PR

csm_web/scheduler/models.py
```diff
@@ -258,7 +259,7 @@ def clean(self):
         if self.spacetime == self.overriden_spacetime:
             raise ValidationError("A spacetime cannot override itself")
         if self.spacetime.day_of_week != self.date.strftime("%a")[:3]:
-            raise ValidationError("Day of week of spacetime and day of week of date do not matchf")
+            raise ValidationError("Day of week of spacetime and day of week of date do not match")
```
Svetlitski: Again, typo

csm_web/scheduler/tests/test_models.py
```diff
@@ -36,7 +34,7 @@ def test_spacetime_with_expired_override(self):
         override = OverrideFactory.create(
             overriden_spacetime=self.spacetime,
             date=date,
-            spacetime=SpacetimeFactory.create(_day_of_week=DAYS[date.weekday()])
```
Svetlitski: DAYS no longer exists after #147

csm_web/scheduler/views.py
```diff
@@ -73,7 +73,7 @@ def get_queryset(self):
     @action(detail=True)
     def sections(self, request, pk=None):
         course = get_object_or_error(self.get_queryset(), pk=pk)
-        #TODO: Clean this up
+        # TODO: Clean this up
```
Svetlitski: Thanks autopep8

csm_web/scheduler/views.py
```diff
@@ -170,13 +170,13 @@ def attendances(self, request, pk=None):
 
 
 class ProfileViewSet(*viewset_with('list')):
-    queryset = None
     serializer_class = None
 
+    def get_queryset(self):
+        pass
+
```
Svetlitski: For some strange reason DRF complaining with an exception when `queryset = None`, but this used to work. I'm guessing updating DRF caused this new behavior, but this little hack fixes it.

csm_web/scheduler/factories.py
```diff
@@ -16,6 +16,8 @@
     Override,
 )
 
+COMPSCI_WORDS = ('Algorithms', 'Systems', 'Distributed', 'Efficient', 'Tractable', 'Programming', 'Languages')
```
noloerino: why don't you throw in machine learning/blockchain while you're at it

csm_web/scheduler/factories.py
```diff
@@ -142,13 +149,22 @@ def demoify_user(user, username):
...
+    demo_mentor_2.user = demo_mentor.user
+    demo_mentor_2.save()
+    demoify_user(demo_mentor.user, "demo_user")
+    demo_student = random.choice(Student.objects.exclude(section__course=demo_mentor.section.course))
+    demo_student.user = demo_mentor.user
+    demo_student.save()
+    print("""
+    A demo account has been created with username 'demo_mentor' and password 'pass'
+    Log in at localhost:8000/admin/
+    """
```
noloerino: This will print the message with one indent unless you call [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent). Just checking if this indent is your intent.

csm_web/scheduler/views.py
noloerino: wouldn't setting `queryset = EmptyQuerySet()` be more idiomatic?

csm_web/scheduler/views.py
```diff
@@ -170,13 +170,13 @@ def attendances(self, request, pk=None):
...
     def list(self, request):
-        student_profiles = StudentSerializer(request.user.student_set.filter(active=True), many=True).data
-        mentor_profiles = MentorSerializer(request.user.mentor_set.exclude(section=None), many=True).data
-        return Response({'mentor_profiles': mentor_profiles, 'student_profiles': student_profiles})
+        return Response(ProfileSerializer([*request.user.student_set.filter(active=True), *request.user.mentor_set.exclude(section=None)], many=True).data)
```
noloerino: I know we can't use the queryset `union` method because they're from two different tables, but this method of concatenating two different querysets feels weird. Do you know if the + operator or something similar can be used here instead of a spread?

Svetlitski: We could use `itertools.chain` to combine to make a single iterator over the two querysets which is passed to `ProfileSerializer`.

csm_web/scheduler/factories.py
Svetlitski: I didn't realize this would have an indent, but I also don't mind it because now that I look at it it does set this off from the progress messages. Will fix only if you want me to.

noloerino: I think the indents fine since it does make it more distinct. Let's leave it as is then.

csm_web/scheduler/views.py
Svetlitski: Didn't know `EmptyQuerySet` existed, will change

csm_web/scheduler/factories.py
Svetlitski: Will fix :)

csm_web/scheduler/views.py
noloerino: I guess the current way is more readable, so we can just keep it as is.

PR #152, opened 2020-01-07
Update RDS certs for CA-2019

noloerino:
The RDS console told me to update certificates, so I pasted the bundle from https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html.


noloerino:
Given that the only changes to the cert file were additive, it should be fine. It's pretty simple to revert anyway.

PR #153, opened 2020-01-07
Bare minimum profile form

noloerino:
This adds the absolute bare minimum functionality to have Scheduler collect bio responses and drop their profile pics into S3 so we can serve them later. Testing this will take a bit of effort, but here are the steps:
1. `brew install awscli`
2. Get the credentials for the test user from https://console.aws.amazon.com/iam/home?region=us-east-1#/users/csm-web-s3-user?section=security_credentials (you'll need to log in to the CSM AWS account)
3. Run `aws configure set aws_access_key_id <access key>` and `aws configure set aws_secret_access_key <secret>` to add your credentials
4. Rerun `./setup.sh` to get the stuff in your `.env`, and reactivate your virtualenv
5. Log in to Scheduler
6. Navigate to `localhost:8000/api/mentor-info` (routing should be changed later since this isn't API)
7. Select the username of the logged in user (it will work fine with any user, but it will only show existing data if you select the current user)
8. Fill in the form and select an image
9. After submitting, look in the bucket to see if a new file was made: `https://s3.console.aws.amazon.com/s3/buckets/csm-web-media/sp20/?region=us-east-1&tab=overview`

TODO:
* Split production and local test instances (mocking is possible, but testing the whole workflow is desirable)
* Frontend
* Allow resume uploading as well (minor, but should be pretty trivial)
* Autofill user field
* Make homepage pull from S3 for picture


Svetlitski:
Going to wait on this until after initial deploy of main feature set, for now we'll omit profile pictures

csm_web/csm_web/settings.py
```diff
@@ -163,6 +162,18 @@
...
+# https://docs.djangoproject.com/en/3.0/topics/files/
+MEDIA_URL = "/media/"
+MEDIA_ROOT = os.path.join(BASE_DIR, "media")
+
+# if DJANGO_ENV == PRODUCTION:
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_STORAGE_BUCKET_NAME = os.environ.get("AWS_STORAGE_BUCKET_NAME")
+AWS_SECRET_ACCESS_KEY = os.environ.get("AWS_SECRET_ACCESS_KEY")
+AWS_STORAGE_BUCKET_NAME = os.environ.get("AWS_STORAGE_BUCKET_NAME")
+# Defaults to public_read, which seems to prevent us from uploading to the bucket
```
Svetlitski: Setting Access Control List to `None` scares me, we should definitely look into this more before deploying anything like this.

csm_web/scheduler/models.py
```diff
@@ -15,6 +15,26 @@ def can_enroll_in_course(self, course):
         )
 
 
+def bio_image_upload_path(instance, _filename):
+    # Get the appropriate semester abbreviation
+    today = timezone.now().date()
+    semester = "fa" if today.month > 6 else "sp"
```
Svetlitski:
```suggestion
    return f"{'fa' if today.month > 6 else 'sp'}{today:%y}/{instance.user.pk}"
```

csm_web/scheduler/models.py
```diff
@@ -15,6 +15,26 @@ def can_enroll_in_course(self, course):
...
+    return semester + today.strftime("%y") + "/" + str(instance.user.pk)
+
+
+class MentorBioInfo(models.Model):
+    """
+    Includes biographical information about a mentor. Only one should exist per mentor;
+    edits to the bio should edit the existing instance in the database.
+    """
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
```
Svetlitski: This seems to rely on mentors having `Mentor` objects in the database so you can associate their bio with the appropriate course(s). But I can see several scenarios where this assumption doesn't hold.
1. Content Mentors currently have bios on the website, but are not on Scheduler
2. Senior Mentors who are not also teaching a section
3. Members of Exec who don't have `Coordinator` objects in the DB because they don't need Scheduler access (like the socials chair, EVP/IVP/President potentially depending on how hands-on they are)

You can't just create "dummy profiles" to associate CMs/SMs with a course because they are related to a course via their section, and then we'd have to create "dummy sections" but have some way of identifying them so they are filtered out of our normal API endpoints and that sounds really hacky.

csm_web/scheduler/models.py
```diff
@@ -15,6 +15,26 @@ def can_enroll_in_course(self, course):
...
+    first_name = models.CharField(max_length=100, verbose_name="First Name")
+    last_name = models.CharField(max_length=100, verbose_name="Last Name")
+    biography = models.TextField()
+    pfp = models.ImageField(upload_to=bio_image_upload_path, verbose_name="Profile Picture")
```
Svetlitski:
```suggestion
    profile_pic = models.ImageField(upload_to=bio_image_upload_path, verbose_name="Profile Picture")
```

csm_web/scheduler/models.py
```diff
@@ -15,6 +15,26 @@ def can_enroll_in_course(self, course):
...
+    return semester + today.strftime("%y") + "/" + str(instance.user.pk)
+
+
+class MentorBioInfo(models.Model):
+    """
+    Includes biographical information about a mentor. Only one should exist per mentor;
+    edits to the bio should edit the existing instance in the database.
+    """
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    first_name = models.CharField(max_length=100, verbose_name="First Name")
```
Svetlitski: Why aren't we just using the `first_name` and `last_name` from the associated `user`?

csm_web/scheduler/views.py
```diff
@@ -298,3 +302,30 @@ class UserViewSet(*viewset_with('list')):
...
+    def get(self, request, format=None):
+        try:
+            info = MentorBioInfo.objects.get(user=self.request.user)
+        except MentorBioInfo.DoesNotExist:
+            info = MentorBioInfo()
+        serializer = MentorBioInfoSerializer(info)
+        return Response({'serializer': serializer})
+
+    def post(self, request, format=None):
+        try:
```
Svetlitski: Use `MentorBioInfo.objects.get_or_create` ([relevant Django docs](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#get-or-create)) instead of all this boilerplate.

setup.sh
```diff
@@ -37,6 +37,20 @@ else
...
+if ! command -v aws 1>/dev/null
+then
+	echo "Did not find AWS CLI installation. You will be unable to view/upload profile pictures for users."
+elif ! aws sts get-caller-identity | grep "csm-web-s3-user" 1>/dev/null
+then
+	echo "You are not logged into the AWS CLI with the credentials for 'csm-web-s3-user'. You will be unable to view/upload profile pictures for users."
+	echo "If calling 'aws sts get-caller-identity' prints 'Partial credentials found in env', check your environment variables. It may be overriding your credentials in ~/.aws."
+else
+	echo "AWS_STORAGE_BUCKET_NAME=csm-web-media" >> .env
+	echo 'AWS_ACCESS_KEY_ID=$(aws configure get aws_access_key_id)' >> .env
```
Svetlitski: This is single-quoted, so the `$(...)` will not actually be run, but included verbatim in .env

setup.sh
```diff
@@ -37,6 +37,20 @@ else
...
+	echo 'AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key)' >> .env
```
Svetlitski: Same issue as above

csm_web/scheduler/views.py
```diff
@@ -298,3 +302,30 @@ class UserViewSet(*viewset_with('list')):
 
     def list(self, request):
         return Response(self.queryset.values_list('email', flat=True))
+
+class MentorBioInfoDetail(APIView):
```
Svetlitski: There's probably an easier way to do this with ViewSet for consistency with the rest of the views

PR #154, opened 2020-01-13
Alter spacetime model

Svetlitski:
Get rid of magic `@property` methods and change everything that interfaced with the Spacetime model accordingly. New model is more conducive to the API we'll need for the redesigned section details views.


csm_web/scheduler/models.py
```diff
@@ -238,20 +202,23 @@ def end_time(self):
...
-        self._location = re.sub(self.SPACE_REDUCE_REGEX, ' ', self._location).strip()
+        self.location = re.sub(self.SPACE_REDUCE_REGEX, ' ', self.location).strip()
         super().save(*args, **kwargs)
 
 
 class Override(ValidatingModel):
     # related_name='+' means Django does not create the reverse relation
     spacetime = models.OneToOneField(Spacetime, on_delete=models.CASCADE, related_name="+")
-    overriden_spacetime = models.OneToOneField(Spacetime, on_delete=models.CASCADE)
+    overriden_spacetime = models.OneToOneField(Spacetime, related_name="_override", on_delete=models.CASCADE)
```
noloerino: is there a reason you're keeping the underscore in the name here?

noloerino: never mind, just saw the `hasattr` on `Spacetime`; I think the `is_expired` logic is best handled in the views though

csm_web/scheduler/admin.py
```diff
@@ -583,7 +583,7 @@ class OverrideAdmin(admin.ModelAdmin):
     fields = ("date", "spacetime", "overriden_spacetime")
     ordering = ("-date",)
 
-    list_filter = ("spacetime___day_of_week", "spacetime__section__course")
+    list_filter = ("spacetime__day_of_week", "spacetime__section__course")
```
noloerino: funny how we forgot to change the underscore in the second filter

csm_web/scheduler/models.py
Svetlitski: I'm not _keeping_ the underscore, I'm _adding_ it. `override` used to be a public field, now we have an `override @property` which returns `None` if there is no override or one exists but it has expired. I did this because it simplifies the serialization of an `Override` when serializing a `Spacetime`.

csm_web/scheduler/admin.py
Svetlitski: We didn't forget, the field `section` (no leading underscore) was always public. If you try to refer to a non-existent field in the admin code (or any other code that gets loaded upon running the dev server) Django will complain and not serve you any pages until it's fixed.

PR #155, opened 2020-01-20
Section Details – Student

Svetlitski:
Students can now review when/where their section is, mentor contact information, and browse through their attendance record.


csm_web/frontend/src/components/Course.js
```diff
@@ -87,8 +87,10 @@ class SectionCard extends React.Component {
 
   static propTypes = {
     id: PropTypes.number.isRequired,
-    location: PropTypes.string,
-    time: PropTypes.string.isRequired,
+    spacetime: PropTypes.shape({
```
noloerino: i see a SPACETIME_SHAPE defined in another file, is it possible to reuse that here?

Svetlitski: Good catch, exported from Section.js to avoid this duplication

Svetlitski:
> FYI i'm getting console warnings about the deprecation/renaming of `componentWillReceiveProps` and `componentWillMount`, but it's not necessarily something that needs to be addressed by this PR

Fixed by updating `react-router-dom`

PR #157, opened 2020-01-22
Add endpoint for spacetime editing

noloerino:
Here's an endpoint that lets you edit spacetimes. I wasn't able to test it manually because I ran into some CSRF problems I didn't have time to figure out, so I wrote a test case instead.

A couple other things I noticed while making this:
- Our tests definitely have race conditions somewhere. After deleting the `test_enroll_drop` to see if the other tests passed, some runs would give 4 failures and others 3 without any changes occurring in between.
- DRF's @action annotation allows GET methods by default ("The action decorator will route GET requests by default, but may also accept other HTTP methods by setting the methods argument." https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing). Visiting `/spacetimes/:pk/modify/` or `/spacetimes/:pk/override/` evidences that specifying actions does not remove GET. I haven't looked into how to get it to not route GET; hopefully there's a way.

### DIFF
--- a/csm_web/scheduler/serializers.py
+++ b/csm_web/scheduler/serializers.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from .models import Attendance, Course, Student, Section, Mentor, Override, Spacetime, Profile
 
 
-class SpacetimeSerializer(serializers.ModelSerializer):
+class SpacetimeReadOnlySerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


Svetlitski:
Sweet, looks like you don't even have to manually rebase this. Merge whenever you please.

PR #158, opened 2020-01-24
Mentor Section Details View

Svetlitski:
All functionality for the Mentor section details view *except* for modifying section time and location


csm_web/frontend/src/components/MentorSection.js
```diff
@@ -0,0 +1,328 @@
...
+  Jun: 6,
+  Jul: 7,
+  Aug: 8,
+  Sep: 9,
+  Oct: 10,
+  Nov: 11,
+  Dec: 12
+});
+
+function formatDate(dateString) {
```
noloerino: Where are these date strings being sourced from? Is it possible to find a way to get this with [Date.toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) or similar?

csm_web/frontend/src/components/MentorSection.js
```diff
@@ -0,0 +1,328 @@
...
+              <tbody>
+                {students.map(({ name, id }) => (
+                  <tr key={id}>
+                    <td>{name}</td>
+                  </tr>
+                ))}
+              </tbody>
+            </table>
+          )}
+          {!loaded && <h5>Loading students...</h5>}
```
noloerino: Would probably be better to use a ternary here in case of some weird race condition that causes both the `loaded` and `!loaded` branches to be rendered simultaneously; I know js is inherently single-threaded but I'd rewrite as a ternary just to be safe in case React or the DOM redraw things weirdly.

csm_web/frontend/src/components/MentorSection.js
Svetlitski: They come from our backend's use of Python strftime. Run the following in a Python interpreter for confirmation
```python
from datetime import date
[date(year=2020, month=i, day=1).strftime("%b") for i in range(1,13)]
```

I suggest we leave them hardcoded because they are tied to the backend, not to any source of truth within the JavaScript standard.

csm_web/frontend/src/components/MentorSection.js
Svetlitski: So I was reading this really long blog post by Dan Abramov (one of the lead maintainers of React) on the `useEffect` hook (https://overreacted.io/a-complete-guide-to-useeffect/), and in the course of describing this he makes it very clear the React runtime guarantees any particular state/props values are tied to a _single_ concrete value during any one render. So we'll never see a mix of states. I try to stay away from the ternary operator for rendering when each option is more than a single line because I feel that it harms code readability, because you're hunting for this single tiny ':' to separate the two possible results.

noloerino:
I don't like the idea of auto-scroll either, since it'll interrupt people who are trying to fill in another attendance bubble. I think the confirmation prompt alone should be sufficient.

Svetlitski:
> I don't like the idea of auto-scroll either, since it'll interrupt people who are trying to fill in another attendance bubble. I think the confirmation prompt alone should be sufficient.

I don't think you understand what I mean by "auto-scroll". All I meant was that when you click the "Attendance" link in the sidebar, the page is automatically scrolled so that the "Attendance" title is at the top of the screen and therefore the Save button is in view.

I realized that a confirmation prompt would actually be very difficult to implement. You'd need some way of "trapping" on all links.

Svetlitski:
There I just implemented the auto-scroll in the last commit. Try it out and tell me what you think, it's only like 10 lines of code so if you're still really opposed to it for some reason it's no big deal to get rid of it.

noloerino:
I thought you meant "when a user modifies an attendance, scroll the save button into view". I think this is fine, although the light color of the save button still might make it a bit hard to notice at first

Svetlitski:
Alright, I'll leave it as is for now. If during QA we find people are confused about it or complain about Roster the we'll make those changes. For now, onwards towards feature completeness.

PR #159, opened 2020-01-29
Add ability for mentors to change section time/location temporarily/permanently

Svetlitski:
With this last bit of functionality, we will be feature-complete on the redesign. All that's left after this (besides anything discovered during QA of course) is cleanup!


csm_web/frontend/src/components/MentorSection.js
```diff
@@ -247,7 +259,132 @@ class MentorSectionAttendance extends React.Component {
...
 }
 
+function zeroPadTwoDigit(num) {
+  return num < 10 ? `0${num}` : num;
+}
+
+class SpacetimeEditModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { location: "", day: "Mon", time: "", isPermanent: false, changeDate: "", showSaveSpinner: false };
```
Svetlitski: Default `day` to Monday because that's the value that's selected by default when the modal opens, so it would be very confusing for request to fail because the user did not _explicitly_ select Monday.

noloerino: might want to make a constant called `DEFAULT_DAY` or similar, rather than hardcoding it

csm_web/frontend/src/components/MentorSection.js
```diff
@@ -247,7 +259,132 @@ class MentorSectionAttendance extends React.Component {
...
+
+  handleChange({ target: { name, value } }) {
+    this.setState({ [name]: value });
+  }
+
+  handleSubmit(event) {
+    event.preventDefault();
+    const { closeModal, spacetimeId } = this.props;
+    let { location, day, time, isPermanent, changeDate } = this.state;
+    isPermanent = !!isPermanent;
```
noloerino: in what case would isPermanent be non-boolean?

Svetlitski:
> Code seems mostly fine, but the first thing I did when messing around with the modal was to create an override for a Monday with the date on a Wednesday. The fact that these didn't match meant the request came back with a 400, but this isn't registered to the user in any way. Is there an easy way to force the user to select a week, or perhaps disable the day of week field for overrides?
> 
> A couple other quality of life things that would be quite useful to have in the future:
> 
> * Imminent overridden dates could be displayed on the homepage for convenience
> * Given that we no longer display the override in place of the permanent date, we should consider allowing multiple overrides to exist at once.
> 
> If you have time, a couple brief API test cases might be useful here, since overrides and spacetimes have a decent amount of extra logic surrounding them.

That's a great point about the redundancy of having to specify both the date and the day of the week. I'll start working right now on changing it so that you only have to specify the day of the week if you're modifying it for all sections.

Imminent overrides would be doable, but potentially confusing (at least for the case of the mentor card where you differentiate between different sections for the same course by when they occur). I can see how this would benefit students, but I'm going to hold off on implementing this right now for the sake of wanting to focus on cleanup and warding off potential bugs.

The multiple overrides thing doesn't make much sense to me. The most common use case for overrides is that your are booking a library room and therefore need to change the location, but you can only book a library room a week in advance so there'd be no reason to need multiple overrides for this situation. I can see certain edge-cases popping up where multiple overrides would be useful, but given that would require changing our data model, I'd rather leave things the way they are unless lots of users are asking for this.

csm_web/frontend/src/components/MentorSection.js
Svetlitski: Always, in fact. The `handleChange` function receives `value` from the target element and uses it to update state, and `value` is always treated as a string.

csm_web/frontend/src/components/MentorSection.js
Svetlitski: Actually just added an un-selectable but default shown placeholder option (with text shown as "---") instead of doing this.

Svetlitski:
Just added tests for the override API endpoint

Svetlitski:
> Looks good, but there's actually one thing I forgot to mention (you can just add it in later)—it would be useful for a newly performed override/modify to show up without needing to refresh the page.

Good point, just added this

PR #166, opened 2020-02-13
Coordinator Interface

Svetlitski:
Created a frontend flow for coordinators much like how students/mentors each have a dedicated flow. I was able to reuse a lot of existing code. The following functionality has been added for Coordinators:

- "Coordinator" card on home page for course(s) they are coord for
- Ability to add sections for their course
- Edit time/location/capacity/description for any section in their course
- Add/drop students to/from any section in their course
- Edit student attendance for any section in their course

I have *deliberately* left out the ability for coordinators to set enrollment start/end times, as enrollment requires scaling our servers, I want to make sure coordinators *have* to ask tech to set this.

### DIFF
--- a/csm_web/csm_web/settings.py
+++ b/csm_web/csm_web/settings.py
@@ -197,6 +197,9 @@ REST_FRAMEWORK = {
     ],
     "DEFAULT_PERMISSION_CLASSES": [
         'rest_framework.permissions.IsAuthenticated'
+    ],
+    "DEFAULT_PARSER_CLASSES": [
+        'djangorestframework_camel_case.parser.CamelCaseJSONParser',
     ]
 }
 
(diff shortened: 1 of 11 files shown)


Svetlitski:
- Not sure what you mean on this first point. All of our permissioning is done in views. I must be misunderstanding you somehow. And as for the permanent modifications, that ability was never actually disabled (unless you did it and pushed it straight to production or something).
- Unfortunately the autocomplete is built into the browser (using `<input>`'s `list` attribute with the `<datalist>` tag), so there's no way to customize its behavior.
- Good point, I'll add this
- Also good point, will tweak the styling here

noloerino:
To the first point, I meant in the sense that providing a read-only serializer is implicitly a form of permissioning, because it prevents endpoints that use it from modifying fields inappropriately. I didn't realize that you didn't actually push a permanent modification fix though, I thought based on slack convos (https://csm-sp20.slack.com/archives/CTGS2FAMD/p1581468119012600?thread_ts=1581463511.011600&cid=CTGS2FAMD) that it was an actual technical fix rather than a "between chair and keyboard"-type fix.

Svetlitski:
Oh I see. The default, accepted pattern in DRF is for serializers to be read/write though, I think it's fine.

PR #172, opened 2020-02-20
Create attendance on student enrollment

Svetlitski:
Solve the confusion around roster/attendance mismatch by automatically creating an attendance for the given week if a student enrolls in a section that hasn't yet been held this week.


csm_web/scheduler/models.py
```diff
@@ -101,6 +109,22 @@ class Student(Profile):
...
+        """
+        now = timezone.now()
+        this_week = week_bounds(now.date())
+        section_day_num = Spacetime.DayOfWeek.values.index(self.section.spacetime.day_of_week)
+        section_already_held = section_day_num < now.weekday() or (
+            section_day_num == now.weekday() and self.section.spacetime.start_time < now.time())
+        course = self.section.course
+        if self.active and course.section_start <= now.date() < course.valid_until\
+                and not section_already_held and not self.attendance_set.filter(date__range=this_week).exists():
+            Attendance.objects.create(student=self, date=now.date())
```
noloerino: Would be useful to add a log message here in case we need to debug in production

Svetlitski: Added

PR #175, opened 2020-02-29
Ban Students

Svetlitski:
Add capability to ban students from a course. Coordinators are prompted for whether or not they should ban a student upon initiating a drop. A "bulk drop and ban" feature for coordinators will be implemented in a separate PR.


csm_web/scheduler/views.py
```diff
@@ -251,7 +258,7 @@ class ProfileViewSet(*viewset_with('list')):
     queryset = EmptyQuerySet
 
     def list(self, request):
-        return Response(ProfileSerializer([*request.user.student_set.filter(active=True), *request.user.mentor_set.exclude(section=None), *request.user.coordinator_set.all()], many=True).data)
+        return Response(ProfileSerializer([*request.user.student_set.filter(active=True, banned=False), *request.user.mentor_set.exclude(section=None), *request.user.coordinator_set.all()], many=True).data)
```
Svetlitski: Strictly speaking it doesn't make any sense for a student to be active and banned simultaneously, but it's safer to be defensive.

Issue #160, opened 2020-01-29
Set up GitHub Actions for CI


Svetlitski:
Checks should include:
- Running all backend tests
- Python linter (autopep8)
- JavaScript linter (eslint)

PR #179, opened 2020-04-21
Add node linter Github Action

noloerino:
See #160. Still needs some configuration to get ESLint/prettier to look in the right places.

This should probably be merged before #178, given that our django tests don't really work.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-csm_web/frontend/src/index.js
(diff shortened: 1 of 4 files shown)


noloerino:
Eslint/prettier now run on PR, and I actually caught a small error with it. It also seems like we're a little inconsistent with where prop shapes are defined, but that can wait for a later PR.

Svetlitski (approved):
Not familiar with how GitHub workflows config works, so trusting you on that, but everything else looks fine.

Issue #164, opened 2020-02-05
Very minor cosmetic changes on mentor view


noloerino:
A couple small things I noticed from today's release that would be nice to have in the future (will update if I think of any more):
- Make "No students enrolled" a little bit grayer or italicized or something
- Add "No students enrolled" for attendance and roster pages

PR #180, opened 2020-04-21
Autofill spacetime on override modal

noloerino:
Implements one of the changes in #164. I tested this in both 24-hr and 12-hr display in both Chrome and Safari, so hopefully it works.

I noticed that the definition of `SPACETIME_SHAPE` seems to have a lot of fields missing - is this supposed to be the case, or are the fields just initialized late? If the former is the case then I'll fix the fields in this PR as well.

### DIFF
--- a/csm_web/frontend/src/components/MentorSection.js
+++ b/csm_web/frontend/src/components/MentorSection.js
@@ -298,11 +298,14 @@ class SpacetimeEditModal extends React.Component {
   constructor(props) {
     super(props);
     // Time string comes as HH:MM:ss, TimeInput expects HH:MM
-    let timeString = props.defaultSpacetime.startTime;
+    const timeString = props.defaultSpacetime.startTime;
+    // Some extra logic in case the API changes to HH:MM,
+    // in which case split would produce 2 segments instead of 3
+    const sliceIndex = timeString.split(":").length < 3 ? timeString.indexOf(":") : timeString.lastIndexOf(":");
     this.state = {
       location: props.defaultSpacetime.location,
       day: props.defaultSpacetime.dayOfWeek,
-      time: timeString.slice(0, timeString.lastIndexOf(":")),
+      time: timeString.slice(0, sliceIndex),
       isPermanent: false,
       changeDate: "",
       mode: "inperson",
@@ -314,7 +317,6 @@ class SpacetimeEditModal extends React.Component {
 
   static propTypes = {
     closeModal: PropTypes.func.isRequired,
-    spacetimeId: PropTypes.number.isRequired,
     defaultSpacetime: SPACETIME_SHAPE.isRequired,
     reloadSection: PropTypes.func.isRequired
   };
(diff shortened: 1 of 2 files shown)


csm_web/frontend/src/components/MentorSection.js
```diff
@@ -296,10 +297,12 @@ function zeroPadTwoDigit(num) {
 class SpacetimeEditModal extends React.Component {
   constructor(props) {
     super(props);
+    // Time string comes as HH:MM:ss, TimeInput expects HH:MM
+    let timeString = props.defaultSpacetime.startTime;
     this.state = {
-      location: "",
-      day: "",
-      time: "",
```
Svetlitski, line 302: I would just add a comment here for future maintainers that this is a React anti-pattern in general because if props updated the changes would not be reflected in this modal, but that it's okay in this instance because the props won't change underneath us.

csm_web/frontend/src/components/MentorSection.js
```diff
@@ -296,10 +297,12 @@ function zeroPadTwoDigit(num) {
...
+      location: props.defaultSpacetime.location,
+      day: props.defaultSpacetime.dayOfWeek,
+      time: timeString.slice(0, timeString.lastIndexOf(":")),
```
Svetlitski, line 305: Can we make this more defensive to make this correct in the case `props.defaultSpacetime.startTime` starts coming in `HH:MM` format in the future?

csm_web/frontend/src/components/MentorSection.js
```diff
@@ -587,7 +591,12 @@ function MentorSectionInfo({
         </InfoCard>
         <SectionSpacetime spacetime={spacetime} override={override}>
           {showModal === MentorSectionInfo.MODAL_STATES.SPACETIME_EDIT && (
-            <SpacetimeEditModal reloadSection={reloadSection} spacetimeId={spacetime.id} closeModal={closeModal} />
+            <SpacetimeEditModal
+              reloadSection={reloadSection}
+              spacetimeId={spacetime.id}
```
Svetlitski, line 596: No need to pass `spacetimeId` as a separate prop anymore if the whole `spacetime` is being passed in (this is just refactoring, what you have works).

csm_web/frontend/src/components/MentorSection.js
```diff
@@ -296,10 +297,12 @@ function zeroPadTwoDigit(num) {
 class SpacetimeEditModal extends React.Component {
   constructor(props) {
     super(props);
+    // Time string comes as HH:MM:ss, TimeInput expects HH:MM
+    let timeString = props.defaultSpacetime.startTime;
```
Svetlitski, line 301:
```suggestion
    const timeString = props.defaultSpacetime.startTime;
```
Arguably bikeshedding on my part, but it never changes and the existing code tends to prefer `const`

noloerino: was probably thinking about rust, where let is immutable by default

csm_web/frontend/src/components/MentorSection.js
noloerino: I'd say that if the API becomes HH:MM we should probably be double checking the frontend consumer of the API anyway, but I've added a check

Svetlitski (approved):
Thanks for addressing comments. Just pulled this and tested it locally, everything works great.

PR #181, opened 2020-06-05
Bump Django to 3.0.7

noloerino:
addresses dependabot alert for django

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ certifi==2019.6.16
 chardet==3.0.4
 defusedxml==0.6.0
 dj-database-url==0.5.0
-Django==3.0.3
+Django==3.0.7
 django-csp==3.6
 django-extensions==2.2.8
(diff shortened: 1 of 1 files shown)


Svetlitski:
What's up with all the new dependencies?

noloerino:
oops, looks like I was using the same venv as for when I implemented #153 - guess this is why we have code review

FYI when I tried to make a new venv from scratch I ran into a psycopg2 compilation error somewhere, fix documented here: https://stackoverflow.com/questions/9678408/cant-install-psycopg2-with-pip-in-virtualenv-on-mac-os-x-10-7?lq=1
export LDFLAGS='-L/usr/local/opt/openssl/lib -L/usr/local/opt/readline/lib -Wl,-dead_strip_dylibs'

Svetlitski:
Looks good now. Yeah the `LDFLAGS` thing is in `setup.sh`, but I should probably document it in the `README` as well.

PR #182, opened 2020-09-04
Support double spacetimes for sections

noloerino:
(PR not fully ready, putting this up just for discussion)

There's two components to this PR:
1. It raises the probability of our tests running successfully from 5% to about 70%. We're still getting random failures due to overrides and other datetime generations producing dates where, say, the override isn't returned by the API because it occurred in the past or occurred too far in the future. This might also mean that our tests will fail more often later in the week if we ever try to set up CI runs without fixing these, since test fixtures use `timezone.now()`.
These errors are fixable, but not a priority at the moment.
2. Adds basic support for sections having 2 spacetimes per week. I implemented the following aspects:
- Changes to test fixtures to support 2x spacetimes
- Display both spacetimes in the mentor/student views for a section
- Display both spacetimes in section signup cards

Remaining TODOS:
- [ ] Squash migrations
- [ ] Test section data entry
- [ ] On course page, allow grouping sections by day of week. There's two slight complications to implementing this. First, on the API side, we're grouping sections by day of week, but it's done by a DB query so implementing this for multiple days of week will be a little bit tricky. Second, 61A (and possibly other classes) are having EOP sections meet twice a week, so we need to be able to group both single day of week and double day of week.
- [ ] Support overrides for 2x spacetime sections: perhaps adding a `second_spacetime` field to overrides as well is the best way to go
- [ ] Nicely format the 2x spacetimes (I just threw in some `<br>`s and stuff)


noloerino:
Also would like to add that CS70's Scheduler Pro (can't find the repo link at the moment) previously supported the 2x spacetime by modifying the day of week enum, under the assumption that a section occurs on two days of the week at the same time both days. This solution could probably work for us as well, but becomes a little inflexible wrt setting overrides, e.g. if a section is regularly M/W @ 1PM but there's a one-time conflict for W only, the enum fix doesn't accommodate this.

Svetlitski:
I'm leaving this open because I'm assuming there's stuff from here you'll want to salvage that wasn't covered in #183

noloerino:
yeah i'll close this myself once I get the test fixes in

noloerino:
addressed by #183

PR #184, opened 2020-09-09
Fix regression preventing students from dropping sections, add placeholder for mentor page

Svetlitski:


Svetlitski:
Checks are not passing because of all the unused variables due to me commenting out most of `<MentorSection>`, so just ignore that.

noloerino:
student card needs to be updated to display both spacetimes if multiple to avoid confusing people (since that's their primary source of truth once they've hit enroll) - otherwise lgtm

Svetlitski:
> student card needs to be updated to display both spacetimes if multiple to avoid confusing people (since that's their primary source of truth once they've hit enroll) - otherwise lgtm

@noloerino It already does that, are you not seeing it locally?
Screen Shot 2020-09-08 at 10 56 34 PM

noloerino:
oh I actually am, my brain just got confused because I thought the middle card was for something else - it's not the greatest design, but it works for now

PR #186, opened 2020-09-14
Multiple sections support for Mentor and Coord views

Svetlitski:
Covers most of the important bases
- Mentors can change section time/location permanently or temporarily
- Mentors can take attendance
- Coords can add/drop/ban students from sections in their course

Huge limitation: attendance display is still broken on the frontend for sections that meet more than once a week (because attendances are still grouped by `weekStart`). Attendance works, but you can't record it for both sections separately.. Because I'm low-key dying right now, I'm not lumping that into this PR. I figure I have until at least Wednesday (when the 2nd section of Monday/Wednesday sections meet) to fix this.


csm_web/frontend/src/components/Section.js
```diff
@@ -108,7 +108,9 @@ export function SectionSpacetime({ spacetime: { location, time }, override, chil
 SectionSpacetime.propTypes = {
   spacetime: SPACETIME_SHAPE.isRequired,
   override: PropTypes.shape({ spacetime: SPACETIME_SHAPE.isRequired, date: PropTypes.string.isRequired }),
-  children: PropTypes.node
+  children: PropTypes.node,
+  manySpacetimes: PropTypes.bool.isRequired,
+  index: PropTypes.number.isRequired
```
noloerino, line 113: maybe add a comment for posterity clarifying that this is 0-indexed but the frontend displays 1-indexed

csm_web/frontend/src/utils/types.js
```diff
@@ -1,9 +1,23 @@
 import PropTypes from "prop-types";
+export const OVERRIDE_SHAPE = PropTypes.shape({
+  date: PropTypes.string.isRequired,
+  // Hacky solution to avoid cyclic dependence between OVERRIDE_SHAPE and SPACETIME_SHAPE
```
noloerino, line 4: better solution would probably be to factor `dayOfWeek`, ..., `time` into a separate shape and pass it as a separate object, but that second layer of indirection would be annoying so this is probably fine as is

noloerino (approved):
overall looks good (tested simple enroll/attendance/drop workflows for student, mentor, and coord and nothing broke), just a few very minor things

csm_web/frontend/src/utils/types.js
Svetlitski: Yeah, I'll leave this for now because I'm in a rush

PR #190, opened 2020-09-17
Restore ability for Coordinators to create sections

Svetlitski:
Coordinators can now easily create sections again, now with support for multiple spacetimes per section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ If you're unfamiliar with CSM and/or its web applications, check out [this repos
 ### Requirements
 We don't know what specific minimum version you would need for any of the following software, but the most recent version of any of the below should work.
 
-* Python 3.7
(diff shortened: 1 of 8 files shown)


noloerino (approved):
lgtm, couple small things though (up to you if you want to fix)

- the x to close the modal gets hidden if you scroll down the modal
- attendances for the current week don't get auto-generated (but i don't think it used to either)

Svetlitski:
> lgtm, couple small things though (up to you if you want to fix)
> 
> * the x to close the modal gets hidden if you scroll down the modal
> * attendances for the current week don't get auto-generated (but i don't think it used to either)

Fixed the first point, good catch

For the second, that's what `manage patchattendances` is for (which I still need to adapt to multiple spacetimes per section).

PR #196, opened 2021-01-30
Make section enrollment list responsive

shadaj:
Before:
Screen Shot 2021-01-30 at 1 53 25 PM

After:
Screen Shot 2021-01-30 at 1 52 40 PM

Also makes test data generation a bit more realistic by only generating weekday sections and creating more per course.

### DIFF
--- a/csm_web/frontend/static/frontend/css/style.css
+++ b/csm_web/frontend/static/frontend/css/style.css
@@ -318,11 +318,12 @@ header {
 }
 
 .section-card {
+	align-self: flex-start;
 	width: 320px;
 	overflow-x: hidden;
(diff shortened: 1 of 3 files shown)


shadaj:
(build failure seems to be an issue with reporting lint results since my account does not have permission to set checks on this repo)

noloerino:
I definitely recall some courses having online sections on weekends in the past, and it'd be good to have some in the test data (especially on Sunday) to make sure our week boundaries are set correctly. Can you revert that change?

shadaj:
@noloerino done!

smartspot2:
Closing as this was resolved in a later commit (a while ago at this point).

PR #197, opened 2021-01-30
Add dark-mode option under a toggle button

shadaj:
Example pages:
Screen Shot 2021-01-30 at 2 35 35 PM
Screen Shot 2021-01-30 at 2 35 41 PM
Screen Shot 2021-01-30 at 2 36 01 PM

The colors for regular mode were also tweaked a bit to use global CSS variables instead of local constants:
Screen Shot 2021-01-30 at 2 35 17 PM
Screen Shot 2021-01-30 at 2 35 27 PM
Screen Shot 2021-01-30 at 2 35 51 PM

### DIFF
--- a/csm_web/frontend/src/components/App.js
+++ b/csm_web/frontend/src/components/App.js
@@ -9,7 +9,10 @@ import LogoNoText from "../../static/frontend/img/logo_no_text.svg";
 import LogOutIcon from "../../static/frontend/img/log_out.svg";
 
 export default class App extends React.Component {
-  state = { error: null };
+  state = {
+    error: null,
+    isDarkMode: window.localStorage.getItem("darkMode") == "true"
+  };
 
   static getDerivedStateFromError(error) {
     return { error };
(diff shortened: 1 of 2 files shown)


csm_web/frontend/src/components/App.js
```diff
@@ -40,13 +40,32 @@ export default class App extends React.Component {
   }
 }
 
+const applyDarkMode = isDarkMode => {
+  window.localStorage.setItem("darkMode", isDarkMode);
+  document.getElementById("app").className = isDarkMode ? "dark-mode" : "";
+  document.body.style.backgroundColor = isDarkMode ? "#1c1d1e" : "white";
+};
+
```
rodsouza9, line 48: this is a pretty hacky way to do this. a very non-react way to add dark mode. look up dark mode with react and change your implementation to follow one of those tutorials

csm_web/frontend/src/components/App.js
```diff
@@ -40,13 +40,32 @@ export default class App extends React.Component {
...
 function Header() {
+  const [isDarkMode, setDarkMode] = React.useState(window.localStorage.getItem("darkMode") == "true");
```
rodsouza9, line 50: same as above. dont do window.localstorage stuff

rodsouza9:
you can no longer see shadow boxes and filled in sections blend in w the background. they should still be noticeably different from the background

csm_web/frontend/src/components/App.js
shadaj: Resolved in latest commit by pulling theme classes into the React tree

csm_web/frontend/src/components/App.js
shadaj: This is required to persist the theme across reloads.

shadaj:
The latest commit adjusts the colors a bit to make disabled cards clear

Screen Shot 2021-02-03 at 5 36 46 PM

PR #201, opened 2021-03-26
Add button to coord interface to download csv of emails of students

rodsouza9:
For now, we are allowing any coord to download csv of any selection of courses.


csm_web/frontend/src/components/Course.js
```diff
@@ -42,7 +43,8 @@ export default class Course extends React.Component {
       currDayGroup: "",
       showUnavailable: true,
       userIsCoordinator: false,
-      showModal: false
+      showModal: false,
+      studentData: false // !!!!!!! SET TO FALSE LATER !!!!!!s
```
noloerino, line 47: Is this just distinguishing between whether the modal should be exporting student data or creating a section? It feels very weird to have this as global state; at the very least it would be better as an enum (e.g. modalKind = "exportCsv" | "createSection") rather than just a boolean.

csm_web/scheduler/views.py
```diff
@@ -110,6 +112,25 @@ def sections(self, request, pk=None):
...
+        id_str = self.request.query_params.get('ids')
+        if not id_str or id_str == "/":
+            return Response({"students": None})
+        if id_str[-1] == "/":
+            id_str = id_str[:-1]
+        ids = id_str.split(",")
+        studs = Student.objects.select_related("user").filter(active=True, section__course__in=ids)
+
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="somefilename.csv"'
```
noloerino, line 127: Make the filename something more meaningful (e.g. "csm-cs61a-student-emails.csv")

noloerino (changes requested):
Overall looks good, a couple usability comments:
- It makes more sense for the "export students" button to download all students for a particular class, since it's accessed through the coordinator interface for a single class - the modal is nice for flexibility though (I'm not too worried about coords downloading student data for other courses either)
- I'm not sure what exactly prompted this feature request, but getting additional information like student name, course, and mentor would be helpful (unless this is explicitly just to get an email list)
  - Also consider using [DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter) to attach header names if you do want to add more fields

csm_web/scheduler/views.py
rodsouza9: I don't want to include course names in the filename cause if the coord selects a large set of courses the name will be annoying.

noloerino: Perhaps just "scheduler-emails-export.csv" or something course-agnostic, but still meaningful.

rodsouza9: yeah I changed it to "csm-student-emails.csv"

noloerino (approved):

PR #202, opened 2021-05-21
Add all deps from pip's requirements.txt to poetry.

rodsouza9:
Due to pip not differentiating between top level and sub deps, we could not maintain this notion in poetry. 

Did my best (ie some deps are saved as sub deps), but most sub deps are saved as top level deps. this shouldn't make a difference in the long run.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,880 @@
+[[package]]
+name = "ansimarkup"
+version = "1.4.0"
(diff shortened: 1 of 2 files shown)


rodsouza9:
need to rewrite setup bash script to handle poetry

smartspot2:
Implemented in #361

PR #203, opened 2021-07-17
Frontend user utils

smartspot2:
Installed typescript dependencies to the project. Also created a new function `getRoles` to fetch user role information from the api.


smartspot2:
Obsolete now; changes have already been reflected in master

PR #208, opened 2021-07-31
Resource agg frontend

Cindyzhang977:
Screen Shot 2021-07-30 at 8 48 13 PM

### DIFF
--- a/csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
+++ b/csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+
+const ResourceEdit = ({ resource , handleChange, handleFileChange, handleSubmit}) => {
+  return (
+    <div className="resourceEditContainer">
+      <div id="resourceEditInner">
+        <div>
+          <div className="resourceInfoEdit">
+            <div>Week Number</div>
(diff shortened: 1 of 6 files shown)


csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+const ResourceEdit = ({ resource , handleChange, handleFileChange, handleSubmit}) => {
```
rodsouza9, line 3: these props need to be typed and put in a ResourceEditProp interface

csm_web/frontend/src/components/resource_aggregation/ResourceWrapper.tsx
```diff
@@ -1,5 +1,16 @@
+import { resourceUsage } from "process";
 import React, { useState, useEffect } from "react";
+import { render } from "react-dom";
```
smartspot2, line 3: I don't think that `resourceUsage` and `render` are used anywhere here?

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -11,6 +11,10 @@ body {
 	overflow-x: hidden;
 }
 
+button:hover {
```
smartspot2, line 14: These CSS edits should also be moved to a new CSS file (created again in the `resource-aggregation-tabs` branch).

PR #215, opened 2021-08-09
Resource aggregation UI

Cindyzhang977:
Screen Shot 2021-08-09 at 12 01 09 AM
Screen Shot 2021-08-09 at 12 01 26 AM


smartspot2:
I'm not too sure when the labels got removed but I think the edit modal should still have labels for the input fields to display what exactly each field is; in my opinion it's quite unclear at the moment.

PR #223, opened 2021-09-05
Add cron jobs with APScheduler

smartspot2:
Created a new process to run a scheduler for cron jobs, replacing Advanced Scheduler from Heroku.

The scheduler runs `create_attendances` every week, and also cleans past execution logs every week.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: cd csm_web; gunicorn csm_web.wsgi
+clock: python csm_web/manage.py run_apscheduler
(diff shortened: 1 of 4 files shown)


smartspot2:
One thing I realized while doing a little more research and deploying to staging is that this scheduler would need to run on an individual dyno on Heroku, and would be running 24/7. I believe that this means that this solution would cost more (unsure of how dyno hours are priced on Heroku). Alternatives would likely be better, especially if all we're doing is scheduling a cron job to create new attendances.

Currently the staging dyno for the clock process is disabled, as I don't want to potentially incur additional charges for something we aren't actually using.

PR #233, opened 2021-09-13
fix attendance generation

rodsouza9:


smartspot2:
FYI: Force push for rebase onto current master state.

smartspot2 (approved):

PR #234, opened 2021-09-16
Add additional validation when a coordinator adds a new student

### LINKED ISSUE #189: Coord student-add function should account for students already enrolled in the course
Right now if a coord goes to add a student to a section, if that student is already in a different section for that course, the add will fail with no explanation on the frontend. The failure is caused by the one-enrollment-per-course restriction, but the problem is the only present way coords can remedy this is by looking through every single section manually to see which section the student is presently enrolled in, so they can drop them. Obviously this is not okay.

smartspot2:
Fixes #189.

Existing errors when coordinators add students to sections occurred if the student already exists in a different section. This PR adds more validation when a coordinator requests to add a student, and adds new frontend UI to display any backend validation errors that occur.

Relevant sentry links:
https://sentry.io/organizations/csm-web/issues/2205103678/events/48376d4acd134bb9acc2a394be3bdfbd/?project=1391866
https://sentry.io/organizations/csm-web/issues/2469049419/events/367953603aed467d91e5ab9e4d4c9acf/?project=1391866


smartspot2:
See #235 for newly merged changes for the mentor meta card.

smartspot2:
Here are some next steps (from comments by Rohan)

Student conflicts:
> But for now, we should handle this by checking, for the student in question (ie the one being added by coord), is there an inactive / active profile that already exists associated with that course. If active, notify the coord that this action will result dropping the student from the other section and to make sure that the student is okay with that. If inactive, check if they are banned. If they are banned, then notify the coordinator that this student has been previously banned and to confirm they want to unban them, then unban them and add them to the new section. If not banned, then just add them with no confirmation check.
So with this flow, sometime the coordinator would need to submit a list and then get info back from the db that makes them check a box and confirm again

Section capacities:
> essentially, the add new students should now take in a list of student emails, then in a modal pop up should show what the new capacity is, give a note that this action will increase the capacity of the section to the new one, and have a confirmation check mark that the coordinator has to check before clicking submit

smartspot2:
Here's the current state of the frontend UI when warnings occur:
image

smartspot2:
Updated modal:
image

noortor (approved):

PR #242, opened 2021-10-23
Migrate scheduler to Typescript

smartspot2:
- Made a heck ton of changes to migrate the site and all components to typescript, creating prop/state interfaces for all components
- Moved all section-related components over to a separate folder
- Added several typing libraries
- Created `images.d.ts` to allow for SVG imports within typescript files
- Added `tsconfig.json` to register `.d.ts` file (may change later if there is a way that we can avoid adding new configuration files)

~~This PR was branched off of #234; it's likely that we will modify the base branch afterward when #234 is merged.~~ Base branch has now been changed to master after merging #234.

FYI: the large diff is due to a change in `package-lock.json` version; it was originally unintentionally reverted to version 1 in commit 625f10a, but has now been changed back to version 2.


smartspot2:
Force pushed to rebase onto new `master` after merging #234.

PR #245, opened 2021-11-11
Refactor scheduler views

smartspot2:
Split the current `scheduler/views.py` into a directory of files, where each file contains one class. The `__init__.py` file in this directory then imports all necessary classes, allowing all imports from `views` to stay valid.

Some potential areas of refactoring:
- Add type hints for as many functions and variables as possible
    - Django type hints may be a little bit tricky, and would likely require extra packages


smartspot2:
Force pushed to rebase with new PR review setup.

PR #261, opened 2022-03-05
Enrollment Policies

ericyche:
Finished enrollment policies :)


csm_web/frontend/src/components/Policies.tsx
```diff
@@ -0,0 +1,498 @@
...
+      i1_b: "Affinity Sections",
+      i1_2:
+        " - These are sections designed for students who affiliate with " +
+        "identities and communities underrepresented in the Berkeley EECS " +
+        "community. Types of affinity sections that we have offered " +
+        "previously, or hope to offer, include:",
+      i1_i1: "1.1 Woman affinity sections",
+      i1_i2: "1.2 LGBTQ+ affinity sections",
+      i1_i3: "1.3 Hispanic/Latinx affinity sections",
+      i1_i4: "1.4 Transfer sections",
```
smartspot2, line 375: The numberings here should probably be omitted, and instead replaced by unordered lists in the HTML

csm_web/frontend/src/components/Policies.tsx
```diff
@@ -0,0 +1,498 @@
...
+              </li>
+              <li>
+                <b> {aff.body.i2_b} </b>
+                {aff.body.i2_2}
+              </li>
+              <li>
+                <b> {aff.body.i3_b} </b>
+                {aff.body.i3_2}
+              </li>
+            </ol>
```
smartspot2, line 98: I'd prefer these lists to be unordered lists rather than ordered lists, to fit in more with the rest of the document.

csm_web/frontend/src/components/App.tsx
```diff
@@ -63,7 +65,7 @@ function Header(): React.ReactElement {
    * That is, it should always be active unless we're in a location prefixed by /resources
    */
   const schedulerActive: NavLinkProps["isActive"] = (match, location): boolean => {
-    return !location.pathname.startsWith("/resources");
+    return !(location.pathname.startsWith("/resources") || location.pathname.startsWith("/policies"));
```
smartspot2, line 68: Here, since we're adding more header items, it may be better just to generalize this, and have a constant array of prefixes that we want to exclude

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -116,6 +125,14 @@ header {
...
 	}
 
+	#site-subtitle-link {
+		font-size: 18px;
+		margin-top: auto;
+		margin-bottom: auto;
+		margin-left: auto;
+		margin-right: auto;
+		padding-top: 5px;
+	}
```
smartspot2, line 135: I believe this is duplicated code?

csm_web/frontend/static/frontend/css/policies-styles.css
```diff
@@ -0,0 +1,34 @@
+.sections {
```
smartspot2, line 1: Regarding the class names, it may be better to be a little more specific than `.sections`, as we ideally do not want any class name conflicts in our codebase—perhaps `.policy-sections` or something similar could be good.

csm_web/frontend/src/components/Policies.tsx
```diff
@@ -0,0 +1,498 @@
...
+  body: {
+    p1_1: "The quick brown ",
+    p1_b: "fox",
+    p1_2: " jumps..."
+  }
+
+  This is then encoded into the HTML through the JS injection with React.
+*/
+
+const SECTIONS = {
```
smartspot2, line 251: Perhaps extract this out to another file, along with the documentation comment? It'll likely make it easier to read through, and keep things more modularized.

smartspot2 (changes requested):
In general, this looks really nice! There are a few comments that I made on the changes that I think could be changed, but please do follow up if you disagree!

One last comment (tangentially related) is that since you're already changing the header, one thing that I think would be good is to shrink the header height a little bit (i.e. just decrease the vertical padding), because before we added all the new tabs, the top part of the webpage wasn't supposed to be a header.

Fully fleshing out what the navbar should look like could be its own project, along with other aesthetic UI changes, though, so do feel free to skip this for now, and we can put another task up in the notion for UI redesigns.

smartspot2:
Is there any update on the feedback and/or changes? We'd like to get this merged sooner rather than later.

csm_web/frontend/src/components/Policies.tsx
rothjason: Resolved, removed numbers and added items to an unordered list!

csm_web/frontend/src/components/Policies.tsx
rothjason: Resolved, converted all lists here to ul's

csm_web/frontend/src/components/App.tsx
rothjason: Resolved, changed schedulerActive to loop through links array for scalability.

csm_web/frontend/static/frontend/css/style.css
rothjason: Resolved, removed duplicate styling.

csm_web/frontend/static/frontend/css/policies-styles.css
rothjason: Resolved, changed "sections" class to "policy-secitons".

csm_web/frontend/src/components/Policies.tsx
rothjason: Resolved, moved policies to static folder in data/

csm_web/frontend/static/frontend/css/policies-styles.css
```diff
@@ -0,0 +1,34 @@
...
+	font-size: 1.05em;
+	line-height: 1.5em;
+}
+
+.policy-sections div {
+	padding-bottom: 0.5rem;
+}
+
+.policy-sections h1 {
+	color: #9ae0b3;
```
noortor, line 13: replace with:
color: var(--csm-green);

csm_web/frontend/static/frontend/css/policies-styles.css
```diff
@@ -0,0 +1,34 @@
...
+	/*
+		CSM green, but IDK if I can import vars from other CSS files lol.
+		OG color was: #646464 (feel free to change back, just trying out green!)
+	*/
+}
+
+.policy-sections a {
+	color: #039be5;
```
noortor, line 21: replace this with var as well

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -72,6 +73,14 @@ header {
 	color: #bbbbbb;
 }
 
+#site-subtitle-link {
```
noortor, line 76: Make this a class. Unless there is specific functionality that requires this to be an id, a class should be used.

noortor (changes requested):

csm_web/frontend/static/frontend/css/style.css
rothjason: Done! Also copied some of the code from site-title-link into site-subtitle-link and removed site-title-link references from policies navlink

smartspot2:
Closing in favor of #317.

PR #268, opened 2022-03-26
Feature: Resource Links

64bitpandas:
New resource links feature for scheduler. We've added a new row to the resources table:
image

Links can be edited similarly to worksheets using the 'New resource' button (or the edit button next to each row):
image


smartspot2 (changes requested):

smartspot2:
Overall, things seem to look fine, though some functionality seems to be broken for me (namely anything involving editing or deleting links), so I can't really test too much locally on my end.

I've made a few comments on style and UI, and it does seem like the prettier CI check has failed—make sure that you're running prettier (either through the VSCode extension or manually) and fixing all warnings and errors.

Another thing; please go through all of the new functions and modified functions and update/create docstrings documenting all of the functionality. See the existing docstrings for reference.

csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
```diff
@@ -246,6 +305,17 @@ export const ResourceEdit = ({ resource, onChange, onSubmit, onCancel }: Resourc
     });
   }
 
+  function handleExistingLinkChange(
```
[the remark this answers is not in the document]
noortor: Same here. I'm not seeing the url field in the state getting updated. It remains as "https://link.here/" and shows nothing in the edit field

csm_web/frontend/src/components/resource_aggregation/ResourceLinkEdit.tsx
```diff
@@ -0,0 +1,59 @@
...
+          type="text"
+          defaultValue={link.url.href}
+          placeholder="Link URL"
+          onChange={e => onChange(e, currentId, "url")}
+          onBlur={() => onBlur()}
+        />
+
+        <button onClick={() => onDelete(currentId)} className="deleteWorksheet">
+          <Trash className="icon" />
+        </button>
```
noortor, line 45: The delete button doesn't seem to do anything when I click on it

noortor (changes requested):

noortor:
My comments are in the same ballpark as Alec's. 
I had some suggestions for improving the UI regarding spacing and placement of components. 
I also encountered errors dealing with unhandled invalid url's and not being able to update and delete links

csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
```diff
@@ -183,7 +279,7 @@ export const ResourceEdit = ({ resource, onChange, onSubmit, onCancel }: Resourc
 
     // validate all fields
     if (validate(true)) {
-      onSubmit(e, existingWorksheetMap, newWorksheets);
+      onSubmit(e, existingWorksheetMap, existingLinkMap, newWorksheets, newLinks);
```
noortor, line 282: Currently, submitting a link without https gets past frontend validation but results in a bad request when sent to the backend. Make sure to send the full link to the backend rather than what is just input by the user.

noortor (changes requested):
Overall, the changes look really good. Great job addressing the UI comments, fixing the update and delete functionality. There is still one bug regarding validation with the backend causing update to sometimes fail.

noortor (approved):

smartspot2 (approved):
This is a really small thing, but the validation in the frontend doesn't seem to match the validation in the backend; it looks like the frontend allows for an empty domain (it doesn't seem like the `URL` object doesn't even split the domain?), whereas Django does actually check for that and rejects the URL instead.

This causes an invisible error in the frontend (i.e. there is no feedback in the UI other than the silent failed URL submission), though all URLs except for the invalid one still succeed. It's not really a big issue, and I'm not _too_ sure how one would solve this easiest (ex. just add more validation in the frontend, or add some kind of further feedback if the backend fails, etc.).

This could just be something to be added in a future update to the UI, so I'll still approve these changes regardless. Nice job!

PR #269, opened 2022-04-02
Word of the Day - backend

sebzhao:
Added a word of the day field to section occurrences, a view for mentors to change the word of the day and a view for students to submit attendance by submitting the word of the day for a given section occurrence.


csm_web/scheduler/tests/models/test_word_of_the_day.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+    student = Student.objects.get(user=student_user)
+
+    attendances = Attendance.objects.all().count()
+    attendance = Attendance.objects.filter(sectionOccurrence=sectionOccurrence).first()
+
+
+    submit_attendance_url = reverse("student-submit-attendance", kwargs={"pk": student.pk})
+    print(submit_attendance_url)
```
smartspot2, line 173: There shouldn't be any print statements in the tests.

smartspot2 (changes requested):
Overall, I think the implementation is pretty solid; when the frontend gets implemented, keep in mind that you may need to add additional views if there are other bits of data that the frontend requires.

Some feedback: in general, with final versions of programs, you shouldn't have any extraneous print statements, and you should clean up your comments as well to be as informative as possible.

I've left a few comments on the implementation of the user request, though one thing to keep in mind is that there should probably be a time constraint on when students can submit the word of the day. You can decide when this should be, but it should be reasonable to give students some leeway. Feel free to brainstorm other kinds of constraints and checks that need to be made as well.

There also seems to be some changes that occurred due to git conflicts/something of that sort—please make sure that you always double check any changes you make before committing or pushing. I personally always do a `git status` to double-check which files have been modified, and then do a `git diff` to slowly walk through all of the changes that have been made (if you've staged files already, `git diff --staged` will diff the staged changes). This way, I know exactly what I'm committing, and what changes will be committed; it has caught many instances of forgotten debug prints, TODO comments, unwanted changes, etc. for me in the past.

For the testing, it could be better to have a few more tests (I think there are a few things that haven't been implemented just yet), so while you're waiting for the frontend to finish, it may be good to think of a few more things to test, and to think about edge cases.

All in all, this is really good work; just re-request my review after you've addressed the feedback!

csm_web/scheduler/views/section.py
```diff
@@ -488,3 +487,31 @@ def _student_add(self, request, section):
...
+
+
+    @action(detail=True, methods=["put"])
+    def change_word_of_day(self, request, pk=None):
+        """
+        Request
+            data = {
+                'word_of_the_day': {new word of the day}
+            }
+        """
```
noortor, line 499: I know this is very nitpicky, but change the docstring to the following to make it consistent with other functions:
"""
Updates the word of the day of a section occurrence.

Format of request.
            request.data is a dictionary with
                'word_of_the_day': new word of the day
"""

csm_web/scheduler/views/student.py
```diff
@@ -91,3 +93,40 @@ def attendances(self, request, pk=None):
...
+
+        section_occurrence_time_limit = datetime.datetime.combine(section_occurrence.date, datetime.datetime.min.time(), tzinfo=timezone.now().tzinfo) + datetime.timedelta(days=1)
+        
+        # Reject any requests made after the deadline (midnight the next day)
+        if request_time > section_occurrence_time_limit:
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        # If the attempt at the word of the day is incorrect, deny request
+        if section_occurrence.word_of_the_day.lower() != request.data['attempt'].lower():
+            return Response(status=status.HTTP_403_FORBIDDEN)
```
noortor, line 125: Add messages to these responses explaining what happened (ie "deadline passed" or "incorrect word of the day"). This way, the frontend knows what to tell the user.

noortor (changes requested):

csm_web/scheduler/views/section.py
```diff
@@ -488,3 +487,32 @@ def _student_add(self, request, section):
                 f"<Enrollment:Success> User {log_str(student.user)} enrolled in Section {log_str(section)}"
             )
             return Response({"id": student.id}, status=status.HTTP_201_CREATED)
+
+
+    @action(detail=True, methods=["put"])
+    def change_word_of_day(self, request, pk=None):
```
noortor, line 493: Since this function involves changing the data of SectionOccurrence, I think this should go in a new section_occurrence.py file.

csm_web/scheduler/views/section.py
```diff
@@ -154,7 +154,8 @@ def students(self, request, pk=None):
 
             - coordinator add:
                 request.user: coordinator that wants to add students
-                pk: primary key of section to enroll into
+                pk: primary key of 
+                section to enroll into
```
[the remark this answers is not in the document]
smartspot2: These changes are still an issue in the latest commit.

csm_web/scheduler/views/section.py
smartspot2: Now that I'm looking at the new functions more, I think it'd actually be better to put all of the functions in their own file with their own API endpoint, ex. `/api/wordoftheday/<section_occurrence_pk>/submit` that takes care of both mentor/coord and student requests.

This method would just check if the `request.user` is a mentor/coord for the course (the course can be retrieved from the section occurrence), in which case the PUT request should change the word of the day to the specified one. If the `request.user` is a student of the course, then the PUT request should check the submitted word and update attendance for the student as necessary.

If you'd prefer, you can also split this up into two functions, one for the mentor/coord, and one for the student. In either case, I think it'd be best to move them into their own file, and use the simpler `@api_view` decorator rather than a viewset (especially since these are basically standalone functions with no backing model).

csm_web/scheduler/views/section.py
```diff
@@ -488,3 +487,32 @@ def _student_add(self, request, section):
...
+
+        sectionOccurrence = get_object_or_error(SectionOccurrence.objects.all(), pk=pk)
+
+        #Checks to make sure the user trying to change the word of the day is actually the mentor
+        if not sectionOccurrence.section.mentor.user == self.request.user:
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        #Do not allow for empty word of the days, if not empty then updates the word of the day for the section Occurrence
+        if request.data['word_of_the_day'] != '':
+            sectionOccurrence.word_of_the_day = request.data['word_of_the_day'].lower()
```
smartspot2, line 510: There should probably also be validation that this is indeed a single word; i.e. reject words that have spaces in them (after stripping initial/leading spaces).

smartspot2 (changes requested):
See the unresolved comments for new changes/continued requests for changes. Please click resolve on the conversations when you've resolved them.

One thing is that I'd recommend installing the `black` formatter locally, so that you can easily format the python files before committing; there are some very long lines and some cases where there are way too many empty lines. The formatter can be installed locally with `pip3 install black`, and it can be run via `black filename`; VSCode also supports the `black` formatter natively, if you search for the settings to enable it as default.

csm_web/scheduler/views/student.py
```diff
@@ -91,3 +93,40 @@ def attendances(self, request, pk=None):
...
+                'attempt': string of attempt for word of the day    
+                'section_occurrence': pk for section occurrence for that week
+        """
+        request_time = timezone.now()
+
+        student = get_object_or_error(self.get_queryset(), pk=pk)
+
+        # Internal server error
+        if student.user != request.user:
+            return Response(status=status.HTTP_404_BAD_REQUEST)
```
smartspot2, line 113: This should be `HTTP_400_BAD_REQUEST`, not a 404.

csm_web/scheduler/views/section.py
sebzhao: If that's the case should I just make the new file and then git revert all of the changes to section.py and student.py (just want to make sure I don't mess up the history more)

smartspot2: If by revert you mean reverting to a previous commit entirely, I'd advise not doing that. You can revert the changes in the files manually, but it should be in its own commit; you shouldn't be changing the history at all. In the final rebase, you can always squash the commits together to make it cleaner.

csm_web/scheduler/views/student.py
sebzhao: Sorry, can you clarify how I should add the messages? Based on the django rest fw documentation, I think the only argument that makes sense would be data.

csm_web/scheduler/views/wordoftheday.py
```diff
@@ -0,0 +1,88 @@
...
+        if request.data["word_of_the_day"] == "":
+            raise PermissionDenied(detail="Word of the day must be non empty")
+        elif bool(re.search(r"\s", request.data["word_of_the_day"].strip())):
+            raise PermissionDenied(detail="Word of the day can not contain white space")
+        else:
+            section_occurrence.word_of_the_day = (
+                request.data["word_of_the_day"].strip().lower()
+            )
+            section_occurrence.save()
+            return Response(status=status.HTTP_200_OK)
```
smartspot2, line 88: One thing here is that if the word of the day is _just_ whitespace, then all the checks would pass, but you'd be saving an empty word of the day.

I'd recommend doing the `.strip().lower()` beforehand and storing it as a local variable, and perform all the checks on the local variable. That way, there won't be any issues between checks and saving.

csm_web/scheduler/views/wordoftheday.py
```diff
@@ -0,0 +1,68 @@
...
+from rest_framework.decorators import api_view
+from django.utils import timezone
+import datetime
+import re
+
+from ..models import Student, SectionOccurrence, Attendance
+from .utils import get_object_or_error
+
+@api_view(['PUT'])
+def submit(request, section_occurrence_pk):
```
noortor, line 14: Since word of the day is a field and not a resource itself, rename the file to section_occurrence.py and change this functions' name to submit_word_of_the_day.

noortor (changes requested):

smartspot2:
Merged into main since I'm taking over the integration of frontend/backend of word of the day.

PR #270, opened 2022-04-15
Feat/archiving/main

rsmcconnell:
Changes only to models.py. Includes:

1. `current_semester_sid()` function outside of any class to clean up filters
2. Validating Model now has `past_objects_included`, an alternative to objects, that includes all objects in the db across all semesters in a query. 
3. Semester Model added. Only one field: `sid`. More recent semesters have higher sid's. By default, all queries to other objects will filter for objects only from the most recent a.k.a. current semester. Because of highlight 2 (above), any query can replace `objects` with `past_objects_included` to bypass this filter. (i.e. `Student.objects.all()` -> queryset of current students, `Student.past_objects_included.all()` -> queryset of students from all semesters in db).
4. Course has a `semester` field. All other models are connected to a semester via the course model.
5. Every model has a manager. For example, Attendance has the AttendanceModelManager. These managers enforce the default current semester only query behavior. Student, Mentor, and other models that inherit from profile don't have individual managers; there is just one manager for Profile.

Note: When first implementing these changes, every course will have its `semester` field set to `None`, and queries will error. This will be resolved as soon as a semester object is created for the current semester and each course's semester field is assigned that semester.


csm_web/scheduler/models.py
```diff
@@ -66,6 +72,22 @@ class Meta:
         abstract = True
 
 
+class Semester(ValidatingModel):
```
smartspot2, line 75: I'd also add a field for the user-readable name of the semester, since currently it's only defined by the `sid`, which is not related to the meaning/interpretation of the semester object. This could just be a small `CharField`, or something similar; I'm envisioning it to just hold something like `sp22` or `fa20`, etc. You could even add some validation to ensure that it's of the form `(sp|fa)\d\d` if you'd like.

csm_web/scheduler/serializers.py
```diff
@@ -1,7 +1,7 @@
 from rest_framework import serializers
 from enum import Enum
 from django.utils import timezone
-from .models import Attendance, Course, SectionOccurrence, User, Student, Section, Mentor, Override, Spacetime, Coordinator, DayOfWeekField, Resource, Worksheet
+from .models import Attendance, Course, SectionOccurrence, User, Student, Section, Mentor, Override, Spacetime, Coordinator, DayOfWeekField, Resource, Worksheet, Semester
```
smartspot2, line 4: It seems like the `Semester` model is imported here without ever being used.

csm_web/scheduler/models.py
```diff
@@ -66,6 +72,22 @@ class Meta:
...
+    sid = models.PositiveSmallIntegerField(primary_key=True)
+    objects = models.Manager()
+
+    def __str__(self):
+        return f"{self.sid}"
```
smartspot2, line 80: If you do end up adding the extra description field, that should be added here to describe the semester model a little bit more.

smartspot2 (changes requested):
In all, this looks really good; the model managers look good, and the semester object overall is pretty good (the fields could be tweaked slightly though).

In testing though, I've found a couple bugs. First, the `pytest` tests are failing because the `Course` factory has not been updated with the `Semester` object; this should be changed to ensure that `pytest` is still working. Some additional tests should probably be added to ensure that the distinguishing of semesters is respected with the queries.

Another bug I've found is that in accessing the section from the UI, an internal error occurs when accessing the `Spacetime.objects` attribute. This could be because it hasn't been overridden, though it seems a little bit weird that the base class would stop working; it'd be good to look into more.

In all, make sure that the `pytest` tests are all passing after each commit (at least, after each commit with finished features), to ensure that no regressions happen. This can be done locally with just `pytest csm_web`, and should run relatively quickly too. If possible, adding more tests would be great, to ensure that the functionality is still working as desired, and to check for any edge cases, though I know that's a work in progress still.

package-lock.json
```diff
@@ -1,7634 +1,8 @@
 {
   "name": "csm_web",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
smartspot2, line 4: It seems like you have an outdated version of node, which changes the lockfile version here; update your node and revert these changes.

PR #284, opened 2022-06-14
Jest and Cypress Testing for React

smartspot2:
We have testing for the backend through `pytest`, but no testing of the frontend functionality. Jest is probably one of the better frameworks out there for testing JS code, and also allows for regression testing through snapshots. Additionally, Cypress can be used for end-to-end testing; this PR adds additional scripts to help with the integration testing, setting up the database for each test.

The overall specification for tests is proposed to be:
* Use Jest for frontend logic tests; avoid using snapshots if possible, as most of the same checks are done with Cypress
* Use Cypress for generic flow tests; this should be for actions that users will go through on the website

Unfortunately, the `package-lock.json` changes dominate the diff stats for this PR.

See the "Testing guide" Notion page for a WIP guide for testing.


noortor (approved):

ericyche (approved):
fantastic :ship emoji:

PR #296, opened 2022-07-24
affinity section

austenliao1:
in progress :))


csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+    const newLabel = { ...oldLabel, name: name, description: description, showPopup: showPopup } as Label;
+    let newLabels = [...tempLabels, newLabel];
+    const positiveLabelsSorted = newLabels.filter((label: Label) => label.id >= 0).sort((a, b) => a.id - b.id);
+    const negativeLabelsSorted = newLabels.filter((label: Label) => label.id < 0).sort((a, b) => b.id - a.id);
+    newLabels = [...positiveLabelsSorted, ...negativeLabelsSorted];
+    setCourseLabels(newLabels);
+  };
+
+  const handleNewRow = () => {
+    const course = courseLabels[0].course;
```
noortor, line 60: This line assumes that courseLabels has elements, which is causing an access error when a course has no labels

noortor, line 60: This line assumes that courseLabels has elements, which is causing an access error when a course has no labels

csm_web/frontend/src/components/course/Course.tsx
```diff
@@ -96,9 +99,19 @@ const Course = ({
   /**
    * Render the currently chosen modal.
    */
+
   const renderModal = (): React.ReactElement => {
```
smartspot2, line 103: The added newline here was probably accidental; it's more conventional to keep the docstring directly attached to the method.

csm_web/frontend/src/components/course/DataExportModal.tsx
```diff
@@ -74,7 +74,7 @@ export const DataExportModal = ({ closeModal }: DataExportModalProps): React.Rea
    */
   const renderCheck = (i: number): React.ReactElement => {
     return (
-      <div className="data-export-checkbox">
+      <div className="data-export-checkbox" key={i.toString()}>
```
smartspot2, line 77: There's no need to convert to a string here; a numeric key works fine.

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+  title
+}: ManageLabelsModalProps): React.ReactElement => {
+  /**
+   * Map of course id to course name.
+   */
+  const [courseLabels, setCourseLabels] = useState<Label[]>([]);
+
+  const [currCourseLabels, setCurrCourseLabels] = useState<Label[]>([]);
+
+  const [newLabelKey, setNewLabelKey] = useState(-1);
```
smartspot2, line 33: I don't believe the first docstring is accurate? I'd also recommend adding docstrings for the rest of the states.

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+    event.preventDefault();
+    const currIDs = courseLabels.map((label: Label) => label.id);
+    const prevIDs = currCourseLabels.map((label: Label) => label.id);
+    const deletedIDs = [];
+    for (const id of prevIDs) {
+      if (!currIDs.includes(id)) {
+        deletedIDs.push(id);
+      }
+    }
+    const put_value = {
```
smartspot2, line 78: [nit] JS/TS typically uses `camelCase` for variables, not `snake_case`.

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+        // begining of evenly spaced row
+        <div className="manage-labels-modal-row">
+          <div className="manage-labels-modal-col large-col">
+            {label.name === "" && <div>Name of label</div>}
+            {label.name !== "" && <div>{label.name}</div>}
+          </div>
+          <div className="manage-labels-modal-col large-col">
+            {label.description === "" && <div>Description of label</div>}
+            {label.description !== "" && <div>{label.description}</div>}
+          </div>
```
smartspot2, line 165: For these, the `=== ""` and `!== ""` lines can be combined using a ternary statement (i.e. `label.name === "" ? ... : ...`)

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+            <input type="checkbox" defaultChecked={label.showPopup} onChange={handleCheck} disabled={true} />
+          </div>
+          <div className="manage-labels-modal-col small-col">
+            <Pencil className="icon matcher-assignment-section-times-edit-icon" onClick={toggleEditingLabels} />
+            <Trash className="icon matcher-assignment-section-times-edit-icon" onClick={removeLabel} />
+          </div>
+        </div>
+        // end of evenly spaced row
+      )}
+      {/* // if editing labels... */}
```
smartspot2, line 176: [nit] There's a lot of comment syntax here lol; you can remove the inner `//` and it'll work fine.

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -196,9 +196,12 @@ header {
 	color: #575757;
 }
 
+.modal-confirmation-container {
+	display: flex;
+}
+
```
smartspot2, line 202: Ideally, your changes to `style.css` should instead be put into its own CSs file, specifically for affinity sections. This is perhaps less relevant to the components you're editing, but for things like the manage labels modal, new CSS styles should always be in their own files, to keep things separate and modular.

csm_web/frontend/src/components/course/CreateSectionModal.tsx
```diff
@@ -104,6 +110,13 @@ export const CreateSectionModal = ({ courseId, closeModal, reloadSections }: Cre
     });
   };
 
+  const handleSelect = (newSelections: any) => {
+    const labelIDs = newSelections.map((label: any) => label.value);
+    setLabels(labelIDs);
+  };
+
+  const animatedComponents = makeAnimated();
```
smartspot2, line 118: What does this do? A comment here explaining a little bit would be useful.

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+    if (type == "name") {
+      onChangeRow(value, label.description, label.showPopup);
+    } else if (type == "description") {
+      onChangeRow(label.name, value, label.showPopup);
+    } else if (type == "showpopup") {
+      onChangeRow(label.name, label.description, value);
+    } else {
+      // should not get here
+      console.error(`Unknown attribute ${type}`);
+    }
```
smartspot2, line 141: This is probably better replaced with a `switch` statement, and the types can also be changed to enums, as it seems like the type string is only used for identification purposes, to distinguish between which fields are being edited.

csm_web/frontend/src/components/course/SectionCard.tsx
```diff
@@ -37,11 +37,41 @@ export const SectionCard = ({
...
+    if (labels.length == 0) {
+      return false;
+    }
+    console.log(
+      labels
+        .map(label => label.showPopup)
+        .some(function (popup) {
+          return popup;
+        })
+    );
```
smartspot2, line 67: This `console.log` should be removed before merging.

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+                    editRow(label.id, name, description, showPopup)
+                  }
+                  removeLabel={() => removeLabel(label.id)}
+                />
+              ))}
+            </ul>
+          </div>
+          <div className="manage-label-modal-buttons">
+            <input className="affinity-btn" type="button" onClick={handleNewRow} value="Add New Label" />
+            <input className="affinity-btn" type="submit" value="Save Labels" />
```
smartspot2, line 109: These buttons seem to be different from other buttons elsewhere; if changes are needed to existing buttons for alignment (which I believe was the cause for the custom class?), the appearance should still be as close as possible to the original.

csm_web/frontend/src/components/course/SectionCard.tsx
```diff
@@ -37,11 +37,41 @@ export const SectionCard = ({
...
+  const [confirmationProcess, setConfirmationProcess] = useState<boolean>(undefined as never);
   /**
    * The error message if the enrollment failed.
    */
   const [errorMessage, setErrorMessage] = useState<string>("");
 
+  /**
+   * Handle the confirmation process in the section.
+   */
+  const confirm = () => {
```
smartspot2, line 52: [nit] The method name is more conventionally something like `handleConfirm` rather than just `confirm`.

csm_web/frontend/src/utils/types.tsx
```diff
@@ -63,3 +63,14 @@ export interface Attendance {
   student: Student;
   presence: string;
 }
+
+// implementing the label type:
```
smartspot2, line 67: This comment should probably be removed before merging.

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -207,9 +210,76 @@ header {
...
+	text-align: center;
+}
+
+.small-col {
+	flex: 1;
+}
+
+.large-col {
+	flex: 7;
+}
```
smartspot2, line 273: These styles seem to be quite generic, and because styles are currently all global, this may cause issues in the future; if they're specifically for the manage labels modal, the classes should be prefixed with something to specify that.

csm_web/scheduler/serializers.py
```diff
@@ -76,10 +76,19 @@ class Meta:
         fields = ("start_time", "day_of_week", "time", "location", "id", "duration", "override")
         read_only_fields = ("time", "id", "override")
 
+# label serializer
+
+
+class LabelSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Label  # ???
```
smartspot2, line 84: The `# ???` should probably be removed before merging.

csm_web/scheduler/tests/models/test_label.py
```diff
@@ -0,0 +1,150 @@
...
+    course = CourseFactory.create(
+        name="course",
+        title="title for course"
+    )
+    mentor = MentorFactory.create(user=mentor_user, course=course)
+    coordinator = CoordinatorFactory.create(user=coordinator_user, course=course)
+    labels = LabelFactory.create(
+
+        course=course
+    )
```
smartspot2, line 58: Reference `batch_create` to create multiple labels at once: https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory.create_batch

The tests should also cover cases where there are no labels.

smartspot2 (changes requested):
It seems like this is still a work in progress, and the error that @noortor pointed out prevents me from testing this locally, so the comments so far are purely from looking at the code, and from a convention/style perspective.

For the frontend, the main thing is to put the CSS styles in its own file, as currently the `styles.css` is _very_ large and does not need to keep growing. Other than that, adding documentation to all the methods would help make the code be more readable, and I believe there are still some comments and logs to clean up (understandable if this is a WIP)

For the backend, it seems like a lot more is commented out compared to frontend, so I'll hold off on comments for now until it's done. The current test is a good start, but do keep in mind that we want to test out all the edge cases as well; think about all the ways in which a user can mess things up, and try to test for them.

Overall, it's quite clear that all of you have put in lots of good work into the project, and it's looking great so far, so keep it up!

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -207,9 +210,76 @@ header {
...
+.affinity-btn {
+	flex-direction: row;
+	font-size: 1.25em;
+	color: white;
+	background-color: var(--csm-green);
+	border-radius: 10px;
+	cursor: pointer;
+	padding: 10px 20px;
+	margin: 0% 2%;
+}
```
noortor, line 224: The button border in the manage label modal doesn't really fit in with the rest of Scheduler, so I'd recommend overriding the default border by setting border-width to 0

csm_web/frontend/static/frontend/css/style.css
```diff
@@ -207,9 +210,76 @@ header {
 	background-color: #007bff;
 	border-radius: 10px;
 	cursor: pointer;
-	left: 0;
-	right: 0;
-	bottom: 20px;
+}
+
+.affinity-btn {
```
noortor, line 215: Since this type of button isn't necessarily just for affinity sections, I think it would be better to name it something more general like `manage-label-btn` or `label-btn` depending on how widely applicable you want this button style to be

csm_web/scheduler/serializers.py
```diff
@@ -76,10 +76,19 @@ class Meta:
         fields = ("start_time", "day_of_week", "time", "location", "id", "duration", "override")
         read_only_fields = ("time", "id", "override")
 
+# label serializer
+
```
noortor, line 80: [nit] remove these lines

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,214 @@
...
+};
+
+interface EditLabelRowProps {
+  label: Label;
+  onChangeRow: (name: string, description: string, showPopup: boolean) => void;
+  removeLabel: (id: number) => void;
+}
+
+export const EditLabelRow = ({ label, onChangeRow, removeLabel }: EditLabelRowProps) => {
+  const [editingLabels, setEditingLabels] = useState<boolean>(false);
```
noortor, line 131: Since only one label is being tracked by this state variable, rename it to "editingLabel"

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,214 @@
...
+  );
+};
+
+interface EditLabelRowProps {
+  label: Label;
+  onChangeRow: (name: string, description: string, showPopup: boolean) => void;
+  removeLabel: (id: number) => void;
+}
+
+export const EditLabelRow = ({ label, onChangeRow, removeLabel }: EditLabelRowProps) => {
```
noortor, line 130: This component manages a label row as a whole rather than just edit functionality, so I think it would be better to name it "LabelRow".

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
```diff
@@ -0,0 +1,214 @@
...
+            {label.description === "" && <div>Description of label</div>}
+            {label.description !== "" && <div>{label.description}</div>}
+          </div>
+          <div className="manage-labels-modal-col small-col">
+            <input type="checkbox" defaultChecked={label.showPopup} onChange={handleCheck} disabled={true} />
+          </div>
+          <div className="manage-labels-modal-col small-col">
+            <Pencil className="icon matcher-assignment-section-times-edit-icon" onClick={toggleEditingLabels} />
+            <Trash className="icon matcher-assignment-section-times-edit-icon" onClick={removeLabel} />
+          </div>
```
noortor, line 179: Add a hover effect for these 2 icons to make it more obvious that they can be interacted with. I would recommend using --csm-neutral for the default icon color and --csm-neutral-darkened when hovered over.

csm_web/frontend/src/components/course/Course.tsx
```diff
@@ -1,9 +1,11 @@
 import React, { useEffect, useState } from "react";
 import { fetchJSON } from "../../utils/api";
 import { Section } from "../../utils/types";
+import { Label as LabelType } from "../../utils/types";
```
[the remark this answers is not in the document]
ashmchiu: fixed!

csm_web/frontend/src/components/course/Course.tsx
ashmchiu: duly noted!

csm_web/frontend/src/components/course/CreateSectionModal.tsx
ashmchiu: done !

csm_web/frontend/src/components/course/ManageLabelsModal.tsx
ashmchiu: donsies !

smartspot2:
@austenliao1 @ashmchiu @yahya-s9 @Veriny Is there any update on this PR? The last changes were from 2 weeks ago.

PR #320, opened 2022-09-20
Coordinator Topic Delimiter Tooltip

edwardneo:
🚢 
Screen Shot 2022-09-19 at 7 33 00 PM

### DIFF
--- a/csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
+++ b/csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
@@ -21,6 +21,7 @@ import CheckCircle from "../../../static/frontend/img/check-circle-solid.svg";
 import ExclamationCircle from "../../../static/frontend/img/exclamation-circle.svg";
 import PlusCircle from "../../../static/frontend/img/plus-circle.svg";
 import ResourceLinkEdit from "./ResourceLinkEdit";
+import { Tooltip } from "../Tooltip";
 
 interface ResourceEditProps {
   resource: Resource;
@@ -693,13 +694,22 @@ export const ResourceEdit = ({ resource, onChange, onSubmit, onCancel }: Resourc
               </div>
               <div className="resourceInfoEdit">
                 <div className="resourceEditHeadItem">Topics</div>
-                <input
-                  type="text"
-                  defaultValue={resource.topics}
-                  placeholder="Topics"
-                  onChange={e => onChange(e, "topics")}
-                  onBlur={() => handleBlur("topics")}
-                />
+                <div className="topicsTooltipWrapper">
+                  <Tooltip
+                    placement="bottom"
+                    source={
+                      <input
+                        type="text"
+                        defaultValue={resource.topics}
+                        placeholder="Topics"
+                        onChange={e => onChange(e, "topics")}
+                        onBlur={() => handleBlur("topics")}
+                      />
+                    }
+                  >
+                    Topics should be delimited by semicolons
+                  </Tooltip>
+                </div>
                 <div className="resourceValidationError">
                   {formErrors.topics && <ExclamationCircle className="icon exclamationIcon" />}
                   {formErrors.topics}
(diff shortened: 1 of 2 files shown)


csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
```diff
@@ -666,40 +667,53 @@ export const ResourceEdit = ({ resource, onChange, onSubmit, onCancel }: Resourc
...
               <div className="resourceInfoEdit">
                 <div className="resourceEditHeadItem">Topics</div>
-                <input
-                  type="text"
-                  defaultValue={resource.topics}
-                  placeholder="Topics"
-                  onChange={e => onChange(e, "topics")}
-                  onBlur={() => handleBlur("topics")}
-                />
+                <div style={{ position: "relative" }}>
```
smartspot2, line 701: More conventionally, put this in a new CSS class with the resources CSS file.

csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
```diff
@@ -666,40 +667,53 @@ export const ResourceEdit = ({ resource, onChange, onSubmit, onCancel }: Resourc
...
-                />
+                <div>
+                  <input
+                    type="text"
+                    defaultValue={resource.weekNum}
+                    placeholder="Week Number"
+                    onChange={e => onChange(e, "weekNum")}
+                    onBlur={() => handleBlur("weekNum")}
+                  />
+                </div>
```
smartspot2, line 678: Are these divs necessary? They don't seem to be doing anything style-wise.

csm_web/frontend/static/frontend/css/resource_aggregation.css
```diff
@@ -214,7 +218,9 @@
 	display: flex;
 	flex-direction: row;
 	justify-content: center;
-	margin: 20px auto;
+	margin-top: 20px auto;
+	margin-left: 20px auto;
+	margin-right: 20px auto;
```
smartspot2, line 223: These can all be combined into one `margin` attribute; the `auto` also doesn't do anything with the CSS if you specify the margin individually.

smartspot2 (changes requested):
In general this all looks fine; just some small CSS changes left.

csm_web/frontend/src/components/resource_aggregation/ResourceEdit.tsx
edwardneo: The tooltip wrapper shifts the entire input down slightly, so we decided the best way to fix this is to add divs to the other two inputs to shift them by the same amount. Is there a better way to do this?

smartspot2: Seems like it's because of this CSS attribute:
https://github.com/csmberkeley/csm_web/blob/fd19c62878643615a83dbbd2d4c1f9c2a15b4461/csm_web/frontend/static/frontend/css/resource_aggregation.css#L229-L231
I'd recommend changing this to instead list out exactly all of the classes that are direct div children of `.resourceInfoEdit`; i.e. I believe it's just `.resourceEditHeadItem` and `.resourceValidationError`, but double check to make sure.

This should mean that you won't need additional divs for the other elements.

csm_web/frontend/static/frontend/css/resource_aggregation.css
```diff
@@ -214,7 +218,7 @@
 	display: flex;
 	flex-direction: row;
 	justify-content: center;
-	margin: 20px auto;
+	margin: 20px 0px 20px 20px auto;
```
smartspot2, line 221: Also here I don't think that `margin` accepts five fields; just omit the `auto` and It'll be good.

smartspot2 (changes requested):
Everything looks pretty good; just one last change with the CSS (for the divs, and for the margin attribute) and I'll approve the PR.

smartspot2 (approved):
Looks good to me; rebase onto master when you can, and we can merge this in.

noortor (approved):

PR #336, opened 2022-11-06
fixed <ul> cannot appear as a descendant of <p> error

### LINKED ISSUE #335: Policies page invalid tag nesting
Simple fix, but there's a rogue `<ul>` tag within a `<p>` tag in the policies page:
image

ericyche:
Screen Shot 2022-11-05 at 9 27 25 PM
No more errors about `ul` being descendent of `p`! 

closes #335

### DIFF
--- a/csm_web/frontend/src/components/Policies.tsx
+++ b/csm_web/frontend/src/components/Policies.tsx
@@ -11,7 +11,6 @@ export default class Policies extends React.Component {
     const aff = SECTIONS.affinitySections;
     const forUnit = SECTIONS.csmForUnit;
     const faq = SECTIONS.faq;
-    console.log(SECTIONS);
     return (
       <div>
         <div className="policy-sections">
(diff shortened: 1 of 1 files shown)


csm_web/frontend/src/components/Policies.tsx
```diff
@@ -177,22 +177,22 @@ export default class Policies extends React.Component {
...
-            <p>
+            <span>
               {faq.body.p8}
               <ul>
                 <li> {faq.body.li1} </li>
                 <li> {faq.body.li2} </li>
               </ul>
-            </p>
-            <p>
+            </span>
```
smartspot2, line 186: Is there any particular reason why the body text needs to be grouped with the list text like this? I feel an easier solution is to just move the closing tag before the `<ul>` tags.

smartspot2:
Additionally, while you're at it, there is a leftover console log as well at the top of the file.

csm_web/frontend/src/components/Policies.tsx
ericyche: oo u are very right! updated

smartspot2 (approved):
Looks good; rebase and squash these commits, and we can merge it.

csm_web/frontend/src/components/Policies.tsx
```diff
@@ -11,7 +11,6 @@ export default class Policies extends React.Component {
     const aff = SECTIONS.affinitySections;
     const forUnit = SECTIONS.csmForUnit;
     const faq = SECTIONS.faq;
-    console.log(SECTIONS);
```
noortor, line 14: squash

noortor (changes requested):

csm_web/frontend/src/components/Policies.tsx
smartspot2: get squashed, kid

noortor (approved):

PR #342, opened 2022-11-16
Update typescript, eslint, prettier, babel dependencies

smartspot2:
Additional typescript-related dependency updates.

Also resolves the following warning when running `eslint`:
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.8.4

Please only submit bug reports when using the officially supported version.

=============
```


smartspot2:
Failing checks are due to new warnings that were not present in the older versions of eslint, prettier, and typescript; we may need to create a new separate PR for resolving these issues before merging this?

Issue #195, opened 2021-01-27
Move from pip to poetry


rodsouza9:
pip is trash
poetry isn't
we need some sort of differentiation between top level dependencies and sub dependencies and dev vs prod deps. 
poetry does that

rodsouza9:
#202

smartspot2:
Closing since this has been implemented a long while ago.

PR #361, opened 2022-12-27
Migrate to Docker and Poetry, add pre-commit CI

smartspot2:
This PR does two things: migrate dependency management to poetry, and migrate all development into docker containers. See #370 for more information about the docker migration; it has been merged into this branch for approval together.

Python dependency management has always been an issue, as we have multiple development dependencies that we don't want to be installed in production. This was originally planned with poetry in #195, and is continued here now that the repository has changed significantly.

Additionally, this PR adds more sophisticated management of pre-commit hooks through the `pre-commit` framework. This means that we have a very simple and easy way to modify and configure pre-commit hooks, ensuring more consistent formatting, and more thorough checks on modified files.

The added format and linter checking does mean that most existing files are not compliant with these styles; to reduce the size of this PR, none of the files will be reformatted in this PR, and should rather be reformatted and refactored when they are next edited. The CI will check formatting and linting on all modified files, and the pre-commit status check will be required for PRs to be merged.

There may be additional PRs later on to do these kinds of formatting and refactoring changes together, so that work isn't unnecessarily put on other projects.


ericyche:
lgtm!

ericyche (approved):
:whale:

edwardneo (changes requested):
I don't know too much about Docker, but on setup I ran into an error about an invalid mount config for type "bind": bind source path does not exist. I fixed it by adding the empty file csm_web/csm_web/frontend/static/frontend/main.js -- do we need to add this file to the branch or make another change to fix this bug?

smartspot2:
> I don't know too much about Docker, but on setup I ran into an error about an invalid mount config for type "bind": bind source path does not exist. I fixed it by adding the empty file csm_web/csm_web/frontend/static/frontend/main.js -- do we need to add this file to the branch or make another change to fix this bug?

@edwardneo  Just made some changes to the node Dockerfile; this is issue due to how Docker volumes work. Apparently, if the file/directory doesn't exist on the host, Docker will error, and existing options in docker-compose always create folders instead of files. I modified the volume mount to instead mount the entire `csm_web/frontend/static/frontend` folder, since that's where webpack outputs the built files. This also led to some changes regarding permissions, which are resolved by adding an entrypoint file for the node container as well.

tl;dr docker is pain, should be fixed

ericyche (dismissed):
xiexie

edwardneo (approved):
Everything seems to work on my end, looks good to me!

ericyche (approved):
yis

PR #368, opened 2023-01-11
Refactor cypress tests to fix retryability flaking

smartspot2:
See https://docs.cypress.io/guides/core-concepts/retry-ability#Only-queries-are-retried.

Existing tests have used `.click().should(...)` to assert for changes after an action, which does not retry the assertion; this means that the corresponding tests are very flaky, especially when state changes across several re-renderings.


ericyche (commented):
refactnoor

ericyche (approved):

noortor:
a

noortor (approved):

PR #369, opened 2023-01-12
Add WebKit support for Cypress tests

smartspot2:
Cypress has some experimental support for WebKit testing, but this requires additional dependencies to be installed, as well as some additional configuration options.

This PR also adds a Cypress CI group to run tests on WebKit on MacOS.


smartspot2:
Upon further research, it seems like service containers are not supported on Windows or MacOS, and can only be used on Linux (https://docs.github.com/en/actions/using-containerized-services/about-service-containers). This means that it's not easy to set up a CI workflow for MacOS cypress tests; the next best solution here is to run the webkit tests on Linux, and hope that there aren't many differences compared to Safari on MacOS.

PR #370, opened 2023-01-12
Migrate to Docker containers

smartspot2:
Setting up the repository has always been a painpoint for onboarding, and using Docker containers rather than doing everything locally seems to be the best option to overcome the differences in machines and environments.

However, local development with Docker would take a lot of research and reading into documentation to setup, so there are a few steps to this migration process:

- [x] Migrate PostgreSQL database to a Docker container
    - The main issue so far has been setting up PostgreSQL on Mac computers, as dependencies have been painful to sort through. It's relatively straightforward to move the PostgreSQL database to run on a Docker container, which would fix this issue.
    - This step would also be necessary for a full migration to Docker anyways, and provides a good stepping stone through the migration process.
- [x] Migrate npm and Django server to a Docker container
    - This would likely be a little more complicated, since it would involve setting up a Dockerfile for the repository for building the project and setting up dependencies.
    - One major thing we want to emphasize is that development should ideally be the exact same process as without Docker: we can make edits to local files, and the Django and npm servers automatically restart with the changes, and these can be accessed directly through a connection from `localhost:8000`.
    - The only part that should be abstracted away is the setup of dependencies; one point to consider is that in development, we still need some development dependencies to have linting, formatting, etc. throughout the editing process, so there would still need to be some local setup involved.
- [ ] Migrate Heroku to a Docker container
    - With production, we do not want to have a Docker container running for the database; we're using AWS RDS for this, so it should not be included in the docker-compose file when we deploy to production. Heroku has very good [Docker support](https://devcenter.heroku.com/categories/deploying-with-docker), so migration should be relatively straightforward, but bugs will almost certainly arise during this process.
    - Edit: We'll likely not use Docker in production at the moment, especially because of the above concerns with bugs and other unforeseen consequences of such a large migration in production. Perhaps later on when scaling is more of a focus, we'll look into Docker containers and Kubernetes clusters, and perhaps also moving toward AWS hosting.

A major priority while doing this migration is to make the development experience as smooth as possible. If it turns out that the migration would cause additional manual work throughout development (ex. we need to manually restart the containers, or there is a lot of duplicated setup in Docker and local environments, etc.), we may scrap this migration in favor of another alternative (at the moment, I don't know what any other alternatives are though).


ericyche:
this works on my machine :holding_back_tears:

ericyche (approved):

smartspot2:
Merging now to collect approvals in #361

Issue #278, opened 2022-06-09
Optimize database queries


smartspot2:
Currently, there is very little to no focus on optimization of database queries. As such, there could be a lot of unnecessary loading/API requests being made to the database from Django, adding to the latency of API responses to the frontend.

To optimize, we should go through the views and add more `prefetch_related` or `select_related` calls, to avoid having to make multiple database requests throughout a view if we know exactly what we'd be referencing in the first place.

See the following stackoverflow question while debugging to see what queries could be optimized: https://stackoverflow.com/questions/4375784/log-all-sql-queries

PR #372, opened 2023-01-15
Optimize test data creation

smartspot2:
Currently, running `createtestdata` takes about a minute to finish, which is quite a long time to wait when not very many objects are being created.

Some optimizations that have been made:
- Use `Factory.build` as much as possible, rather than using `Factory.create`, so that we can save all the instances at once with `bulk_create`. This requires a bit of shuffling around to get dependencies created correctly, especially since running `bulk_create` does not create related objects (ex. foreign key instances created by subfactories).
- Use `select_related` and `prefetch_related` in loops after creation, to avoid large amounts of queries in loops.

Additionally, the `factories.py` file has been reformatted along with these changes. In the future, we can do similar kinds of optimizations (as with #278) for other views.

### DIFF
--- a/csm_web/scheduler/factories.py
+++ b/csm_web/scheduler/factories.py
@@ -27,8 +27,9 @@ from .models import (
 )
 
 
-def evaluate_faker(faker):
-    return faker.evaluate(None, None, {"locale": None})
+def evaluate_faker(faker_obj: factory.Faker):
+    """Evaluate a faker object with all defaults"""
+    return faker_obj.evaluate(None, None, {"locale": None})
 
 
 COMPSCI_WORDS = (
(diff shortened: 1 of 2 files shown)


csm_web/scheduler/factories.py
```diff
@@ -111,9 +113,9 @@ class Meta:
         lambda: evaluate_faker(factory.Faker("name")).split()[-1]
     )
     username = factory.LazyAttributeSequence(
-        lambda o, n: "%s_%s%d" % (o.first_name, o.last_name, n)
+        lambda o, n: f"{o.first_name}_{o.last_name}{n}"
```
ericyche, line 116: wow f string so beautiful

ericyche (approved):
lgtm

PR #376, opened 2023-01-17
Hotfix: Errors when no attendances, Optimize courses endpoint

smartspot2:
This is a PR for several hotfixes for the prod release of CS70 sections:
* There was an existing frontend issue that does not allow users to view attendances if no attendances exist.
* Additionally, some minor adjustments were made to attendance creation commands/views for edge cases.
* The courses endpoint was extremely slow, retrieving a query of over 5 million objects (before the `.distinct()` modifier) when really only around 7 courses are possible. This was optimized and rewritten with more queries but for fewer objects.


ericyche (approved):
looks legit. very cool, very alec

noortor (approved):
a

PR #380, opened 2023-01-25
Prevent mentors from submitting less than 3 nonzero preferences

### LINKED ISSUE #333: Matcher Exploit
The enrollment automation matching algorithm implicitly prioritizes mentors who have limited availability, so they are able to input 1 time and get their preferred slot. We should add validation to ensure that all mentors enter a reasonable amount of availability. This may be done using a per-preference basis, a preference sum, or any other metric.

noortor:
Resolves #333.

Fixes the matcher exploit by posing a hard restriction on the number of nonzero preferences a mentor can submit. Currently, this is a hard-coded limit of 3 slots, but in the future, this should be configurable by the coordinator.


smartspot2 (approved):
Ignoring failing CI action for now; approving to merge for release to prod tomorrow.

ericyche (approved):
very if statement. thank u 101noor

PR #385, opened 2023-02-03
Email Case Sensitivity

edwardneo:
Added lowercase email field to override Django's case sensitive email field


smartspot2 (changes requested):
This looks fine, but there's still a persistent issue where:

1. Student already exists in the database, with a particular email
2. Coordinator adds a student to a section by email, but this email has different case compared to the email of the user in the database
3. Now there's two different users with the same (case-insensitive) email.

PR #393, opened 2023-02-17
fix for issue389

Veriny:
yea


smartspot2 (changes requested):
This seems good overall; could you also create tests (Cypress and pytest) for manually adding students to sections, and attempting to take attendance for them?

PR #398, opened 2023-02-28
Section Swap Backend

edwardneo:


ericyche:
make wording more descriptive -1

edwardneo:
:(

csm_web/scheduler/models.py
```diff
@@ -243,9 +240,18 @@ class Meta:
...
+    a Swap instance is initialized containing references to the respective students.
+    """
+    sender = models.ForeignKey(Student, related_name="sender", on_delete=models.CASCADE)
+    receiver = models.ForeignKey(Student, related_name="receiver", on_delete=models.CASCADE)
+
+
 class Mentor(Profile):
     """
-    Represents a given "instance" of a mentor. Every section a mentor teaches in every course should
+    Represents a given "instance" of a mentor. Every section a mentor teaches in every ccurse should
```
ericyche, line 254: lmao

KartavyaSharma: Fixed.

ericyche: thank you kurt :holding_back_tears:

csm_web/scheduler/tests/models/test_swap.py
```diff
@@ -0,0 +1,233 @@
...
+    # Decode get response to UTF-8
+    swaps = json.loads(swaps.content.decode("utf-8"))
+    if not is_empty:
+        # Make sure that the swap request was created
+        assert Swap.objects.filter(receiver=reciever_student).exists()
+        # Make sure that the swap request is in the list of swaps
+        assert len(swaps["sender"]) != 0 and swaps["sender"][0]["receiver"] == reciever_student.id
+    else:
+        assert len(swaps["sender"]) == 0
+
```
smartspot2, line 76: This should be split up into two tests; the parameterization should be for a single _kind_ of test with various different inputs; here, you're testing two different situations; one where the student has no requests, and one where the student has a swap request.

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+        # check if student_id is specified in request
+        if student_id == -1:
+            raise PermissionDenied("The `student_id` field in the request cannot be empty,"
+                                   "please specify student.")
+
+        # check if student exists
+        try:
+            student = get_object_or_error(Student.objects, id=student_id)
+        except Exception:
+            raise NotFound("No student found with id: " + str(student_id))
```
smartspot2, line 830: Generally you should never be catching the generic `Exception` object, as this means that any unexpected errors would also be caught and we wouldn't know about it. Make all of the `except Exception:` blocks throughout more specific, for _only_ the kinds of exceptions you expect could happen as a result of user error.

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+
+            '''Create a new Swap object between the the current student
+            and the student with the email specified in the request. This is essentially
+            creating a swap invite.'''
+            receiver_email = request.data.get("receiver_email", None)
+
+            # Check if receiver_email is empty
+            if receiver_email is None:
+                raise PermissionDenied("The `receiver_email` field in the request cannot be empty,"
+                                        "please specify a receiver.")
```
smartspot2, line 859: User input errors should be returned as a 400 Bad Request error, not a 403 Permission Denied error (unsure if there's a builtin response class for this). A 403 (raised by `PermissionDenied`) is for when the user is not authorized to access a resource or perform an action.

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+                    )
+                    raise NotFound("Could not find swaps for student.")
+                sender.section = target_swap.receiver.section
+                # Get outgoing and incoming swaps for sender
+                outgoing_swaps = Swap.objects.filter(sender=sender).select_for_update()
+                incoming_swaps = Swap.objects.filter(receiver=sender).select_for_update()
+                for expired_swap in outgoing_swaps:
+                    expired_swap.delete()
+                for expired_swap in incoming_swaps:
+                    expired_swap.delete()
```
smartspot2, line 945: There's a bulk delete method for a `QuerySet`: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#delete

Using a for loop sends separate queries for each delete action, while a delete on a query set will send one query to delete all of the objects in the query set.

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+            if student_id != swap.reciever.id and student_id != swap.sender.id:
+                logger.error(
+                    f"<Swap Deletion:Failure> Could not delete swap, "
+                    f"user {log_str(request.user)} does not have proper permissions"
+                )
+                raise PermissionDenied("You must be a student or a coordinator"
+                                       "to delete this spacetime override!")
+
+            swap.delete()
+            swap.save()
```
smartspot2, line 976: There's no need to save the object if you've just deleted it.

csm_web/scheduler/tests/models/test_swap.py
```diff
@@ -0,0 +1,233 @@
...
+
+def reject_swap_request(client, receiver, swap_id):
+    """
+    Receiver rejects a swap request from a sender.
+    """
+    client.force_login(receiver.user)
+    delete_url = reverse("section-swap-with-id", kwargs={"section_id": receiver.section.id, "swap_id": swap_id})
+    response = client.delete(delete_url)
+    return response
```
smartspot2, line 233: Add a newline at the end of the file >:(

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+            return Response({}, status=status.HTTP_201_CREATED)
+
+    
+    @action(detail=True, methods=["POST", "DELETE"], url_path='swap/(?P<swap_id>[^/.]+)')
+    def swap_with_id(self, request, section_id=None, swap_id=None):
+        '''If swap_id is not None, initiate the swap between the two students. This should 
+        be done in one atomic transaction. If either or both of the swaps are already 
+        deleted, then return a 404 error.'''
+        student_id = request.data.get("student_id", -1) if request.method != "GET" \
+            else request.query_params.get("student_id", -1)
```
smartspot2, line 903: Since this inline conditional is quite long (it spans two lines), it's better convention (and more readable) to split it up into if/else blocks.

This should conventionally default to None if the field is not present, not -1. Additionally, the allowed request methods are only POST and DELETE, so the conditional doesn't quite make sense to me; the `else` part is never executed.

csm_web/scheduler/views/section.py
```diff
@@ -779,3 +784,198 @@ def wotd(self, request, pk=None):
...
+                receiver.section = target_swap.sender.section
+                # Get outgoing and incoming swaps for receiver
+                outgoing_swaps = Swap.objects.filter(sender=receiver).select_for_update()
+                incoming_swaps = Swap.objects.filter(receiver=receiver).select_for_update()
+                for expired_swap in outgoing_swaps:
+                    expired_swap.delete()
+                for expired_swap in incoming_swaps:
+                    expired_swap.delete()
+                receiver.save()
+                # Finall, delete the swap object
```
smartspot2, line 956: typo

smartspot2 (changes requested):
Overall, this looks pretty good; just some comments on convention. I haven't actually run any of the code though; I'll likely do that when this gets merged into the main feature PR, as when integration is done, some bugs are likely to surface.

KartavyaSharma:
@smartspot2 Fixed all the requested changes, do you want me to merge this into the feature branch?

csm_web/scheduler/views/section.py
```diff
@@ -17,19 +19,23 @@
...
+    SwapSerializer,
 )
 
 from .utils import get_object_or_error, log_str, logger, viewset_with
 
 
 class SectionViewSet(*viewset_with("retrieve", "partial_update", "create")):
     serializer_class = SectionSerializer
+    lookup_field = 'pk'
+    lookup_url_kwarg = 'section_id'
```
smartspot2, line 38: These seem to break the existing section tests due to the change in URL matching. I'm unfamiliar with these viewset fields, but this should be looked into; either modify this, or correct the existing tests (and potentially serializers?) that may rely on the URL match groups.

smartspot2 (changes requested):
I've just pointed out a couple more small issues; I've also ran the pytests locally, and they seem to be failing due to the changes in the URL matching for the Section viewset. After you've fixed these, you can merge this into the main feature branch (it doesn't need approval).

When you've merged this in, make sure you also rebase the main feature branch onto master, and ensure that these changes work with the new Django version and with Docker. The CI should automatically run with that PR when this is merged in, so you can fix any additional bugs that may have popped up there too.

csm_web/scheduler/views/section.py
KartavyaSharma: @edwardneo I remember this coming from a commit of yours for the swap DELETE method. I have removed this for now since it's messing up the URL matcher for all the other tests and I was able to get the swap tests to work without this by making a few changes in the parameter names for the `swap_with_id` and `swap_no_id` functions. Lmk if you need me to add this back for any code extensions you plan to make in the future. Leaving this unresolved for now.

KartavyaSharma:
@smartspot2 Should I squash + merge here or rebase + merge, what do we usually do for merging into feature branches?

PR #399, opened 2023-02-28
User Profile Backend Models

adit-bala:


csm_web/scheduler/views/user.py
```diff
@@ -23,12 +23,29 @@ def list(self, request):
...
+        if bio is not None:
+            user.bio = bio
+        if pronunciation is not None:
+            user.pronunciation = pronunciation
+        if pronouns is not None:
+            user.pronouns = pronouns
+        if is_private is not None:
+            user.is_private = is_private
+        user.save()
+        return Response(status=status.HTTP_201_CREATED)
```
smartspot2, line 51: Ideally, this shouldn't be a `201 CREATED`, since you haven't actually created anything. Likely a `200 OK` or a ``202 ACCEPTED` would likely fit better.

adit-bala: yes sir

smartspot2 (changes requested):
The diff is dominated by whitespace (EOL) changes; I'd recommend trying to resolve this before merging, as it'll make the actual diff a lot harder to view on GitHub.

Perhaps look at git documentation on line endings, or look at line ending configuration in your editor, and tweak configuration settings until modifications to files do not cause all line endings to be changed.

PR #413, opened 2023-05-09
Fix export attendance script

smartspot2:
The current `export_attendances` command uses an old version of the attendance model, without section occurrences. It also had an issue where exported attendances would be duplicated (due to an inner join in selecting spacetime objects for ordering), and this has been fixed as well. The file was also reformatted via `black`.


ericyche:
code looks like code that works

ericyche (approved):
very legit

edwardneo (approved):
Reviewed :)

PR #421, opened 2023-07-05
Add `workflow_dispatch` trigger to all actions

smartspot2:
Since the current GitHub actions are triggered by push and PR events on the master branch, any PRs or pushes to other branches do not trigger any workflows, and also cannot be manually triggered either. In case we'd like to run any CI actions on these extraneous branches, adding the `workflow_dispatch` trigger can be helpful, as it allows us to manually trigger workflows from the Actions tab: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -5,6 +5,7 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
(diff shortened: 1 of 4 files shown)


edwardneo (approved):
yay.

PR #423, opened 2023-07-17
Add Luxon, refactor date parsing/formatting

### LINKED ISSUE #367: Add better datetime frontend library
Native JS has horrible support for dates/times, especially with timezones. Since our database stores everything in PST (for convenience, as basically everybody accessing the website is in PST, and international support would make days of the week painful to handle), we need a more robust way to parse, format, and manipulate the raw dates produced by requests.

This is especially evident in the below snippet, taken from #360, where some very cursed date parsing was required to compare two date objects: https://github.com/csmberkeley/csm_web/blob/5b5464db739e34d8b700f56362b866d66fc7d642/csm_web/frontend/src/components/section/StudentSection.tsx#L233-L275
This could all be simplified with the use of a third-party library.

The library currently in mind is [Luxon](https://github.com/moment/luxon/) ([documentation](https://moment.github.io/luxon/#/)), which is essentially a newer version of [Moment.js](https://momentjs.com). The reason why I personally think Luxon would be a better fit compared to Moment.js is due to the shift in focus from parsing to manipulation. Moment is built on the premise that one can parse any datetime representation very easily, whereas Luxon is built more for purposeful actions on strictly formatted datetime strings. I'd be open for discussion on Luxon, Moment.js, or other alternatives though!

smartspot2:
Closes #367

Default datetime support in JS is horrible, so Luxon was added to make it a lot easier to manage.

Further, there are some serializers which format dates in the response, which is bad convention, as the backend should provide dates in a standardized format (i.e. ISO), letting the frontend do the formatting; parsing should be made as easy as possible. This was changed in this PR so that serializers give all date/datetime values in ISO.

### DIFF
--- a/csm_web/frontend/src/components/CourseMenu.tsx
+++ b/csm_web/frontend/src/components/CourseMenu.tsx
@@ -35,7 +35,7 @@ const CourseMenu = () => {
   if (userInfoLoaded) {
     let priorityEnrollment = undefined;
     if (jsonUserInfo.priorityEnrollment) {
-      priorityEnrollment = DateTime.fromISO(jsonUserInfo.priorityEnrollment);
+      priorityEnrollment = DateTime.fromISO(jsonUserInfo.priorityEnrollment, { zone: DEFAULT_TIMEZONE });
     }
     const convertedUserInfo: UserInfo = {
       ...jsonUserInfo,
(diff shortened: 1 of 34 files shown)


ericyche:
o this is cool wow

ericyche:
i test and i like date format.

ericyche (approved):

edwardneo (approved):
good luxon migration :D

PR #427, opened 2023-08-07
Refactor admin pages

### LINKED ISSUE #364: /admin does not take you to django admin site
### How to recreate error: 
Setup local environment and enter `localhost:8000/admin` redirects you to `http://localhost:8000/login/?next=/admin` instead of the django admin page. 
Screen Shot 2023-01-02 at 6 52 34 AM

### Temporary Fix
Instead of `localhost:8000/admin`, add a slash to the end like `localhost:8000/admin/`

### Context
"it used to work fine the way you put it, but when we swapped to a browser router rather than a hash router i had to change some django routing that broke it" - da lao alec

### LINKED ISSUE #384: Admin permissions to delete spacetimes
The admin interface has a lot of legacy code, and currently the `SpacetimeAdmin` overrides the delete permission preventing anybody from deleting spacetimes (and therefore preventing anybody from deleting sections), citing a one-to-one relation as the cause. However, a git blame on the admin file shows that the override was last edited in 82c60ae8, and the `Section` model at the time of that commit had a one-to-one field to `Spacetime`, which is not the case now.

It should be a fairly quick fix to remove the permission override, as it is no longer an issue to delete spacetimes.

smartspot2:
Fixes #364, fixes #384.

The admin pages are all very outdated, and this PR does a lot of refactoring to make the lists more helpful, and fix a lot of legacy code (especially regarding permissions).

This PR also fixes some URL issues when trying to navigate to the admin page. In particular, the existing URL configuration to route to the frontend wasn't optimal; there was a hard-coded check that the URL does not begin with `api/`. This was changed to two routes; one routing from `/` to the frontend, and one routing any path `^.*/$` to the frontend.

The root cause for #364 turns out to be how Django handles URL matching. Since Django automatically adds a trailing slash if a path matches nothing, it is important that _all_ routes end with `/`, so that paths that do not end in a slash get rerouted with the `/` appended. Prior, the frontend catch-all route was for `^(?!api/)`, which _does_ match paths that do not end in a slash. This means that `/admin` will match this URL, and get passed to the frontend, which is undesired behavior.

### DIFF
--- a/csm_web/frontend/urls.py
+++ b/csm_web/frontend/urls.py
@@ -1,4 +1,10 @@
-from django.urls import re_path
+from django.urls import path, re_path
+
 from . import views
 
-urlpatterns = [re_path(r"^(?!api/)", views.index)]
+urlpatterns = [
+    # catch index path
+    path("", views.index),
+    # catch all other subpaths
+    re_path(r"^.*/$", views.index),
+]
--- a/csm_web/scheduler/admin.py
+++ b/csm_web/scheduler/admin.py
@@ -1,27 +1,21 @@
-from django.db.models import Count, Q
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.core.exceptions import ValidationError
-from django.utils.html import format_html
-from django.urls import reverse
 from django.db import transaction
-from django.contrib import messages
-from django import forms
-
+from django.db.models import Q
+from django.urls import reverse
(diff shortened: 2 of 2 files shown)


ericyche (approved):
at very least path thing is fixed on front end. i dont' know the backend too well so i make edward reed it more carefully

edwardneo (approved):
Everything looks good to me!

PR #431, opened 2023-08-28
Update pyproject dependencies for prod shell access

smartspot2:
The update to docker causes issues with accessing the production shell; there are dependencies on `pytest` and `freezegun` which are not satisfied. This PR adds those dependencies to the production section of the pyproject file, so that they are installed in the Heroku instance.


edwardneo (approved):
Looks good to me

ericyche (approved):
god bless

PR #434, opened 2023-09-02
Fix matcher scrolling

smartspot2:
With the luxon migration, the matcher was refactored to use luxon datetime objects, and as a result of this change, the matcher scrolling behavior broke. The cause was a missed `.plus` rather than a `.minus` on a scroll event.


ericyche (approved):
approving out of spite

smartspot2:
#435 forcing my hand at closing this pr smh

PR #438, opened 2023-09-04
Fix multiple mentors added to matcher

ericyche:
Matcher would create multiple mentors with the same user. We turn the input list into a set so there cannot be any duplicate emails. 
https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html more details about flags i added. 
Screenshot 2023-09-03 at 11 11 19 PM
Screenshot 2023-09-03 at 11 10 57 PM

### DIFF
--- a/csm_web/scheduler/views/matcher.py
+++ b/csm_web/scheduler/views/matcher.py
@@ -291,7 +291,7 @@ def mentors(request, pk=None):
         skipped = []
         # users already associated with the course as a mentor
         users_with_course = User.objects.filter(mentor__course=course)
-        for email in request.data["mentors"]:
+        for email in set(request.data["mentors"]):
             if not email or "@" not in email:
                 # invalid or blank email
                 skipped.append(email)
--- a/csm_web/scheduler/views/matcher.py
+++ b/csm_web/scheduler/views/matcher.py
@@ -1,37 +1,34 @@
 import datetime
 
+from django.db import transaction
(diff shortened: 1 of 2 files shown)


ericyche:
i am never safe from pre-commit

csm_web/scheduler/views/matcher.py
```diff
@@ -520,21 +518,24 @@ def run_matcher(course: Course):
         - unmatched: [int, int, ...] of mentor ids
     """
     # get slot information
-    slots = MatcherSlot.objects.filter(matcher=course.matcher)
+    matcher_slots = MatcherSlot.objects.filter(matcher=course.matcher)
     # get preference information
-    preferences = MatcherPreference.objects.filter(slot__matcher=course.matcher)
```
smartspot2, line 525: This should be `matcher_preferences` (plural) to match the other variables (and the semantics of this variable)

smartspot2 (changes requested):
Fix the one variable name and I'll approve the next revision.

ericyche (commented):
idk what this page is on github but i will add a comment.

smartspot2 (changes requested):
Please squash and remove the merge commit.

ericyche:
not fan of merge commit :(

edwardneo (approved):
Nice!

smartspot2 (approved):

PR #439, opened 2023-09-07
Fix change day for section in coord interface

ericyche:
Fixed bug where coordinator was not able to change the day of a recurring seciton. 
Screenshot 2023-09-07 at 12 33 00 AM

PR #440, opened 2023-09-10
Fix section day change form

smartspot2:
See #439 for the original fix of the issue, but after merging in #429, there were additional errors due to the types used in the form (the state change failed to convert a string to an int type). This is a follow-up PR fixing that issue as well.


ericyche (approved):
i see

edwardneo (approved):
Tested and approved!

PR #443, opened 2023-09-13
Add delete button for coordinators to drop sections

KartavyaSharma:


csm_web/frontend/src/components/section/MentorSectionInfo.tsx
```diff
@@ -1,20 +1,37 @@
 import React, { useState } from "react";
-import { useSectionStudents } from "../../utils/queries/sections";
-import { Mentor, Spacetime, Student } from "../../utils/types";
+import { Navigate, Route, Routes } from "react-router-dom";
```
smartspot2, line 2: There are a lot of ESLint warnings here about unused variables; you should delete the unused variables here.

csm_web/frontend/src/css/coordinator-add-student.scss
```diff
@@ -308,3 +308,45 @@ $modal-effective-height: calc($modal-height - $modal-padding-y);
...
+
+  // border: none;
+  // padding: 0;
+  // font: inherit;
+  // cursor: pointer;
+  // outline: inherit;
+
+  // :hover {
+  //   background-color: #f15120;
+  // }
```
smartspot2: If these styles are unused, they should be deleted, rather than commented out.

csm_web/scheduler/views/section.py
```diff
@@ -119,6 +119,43 @@ def create(self, request):
         serializer = self.serializer_class(section)
         return Response(serializer.data, status=status.HTTP_201_CREATED)
 
+    def destroy(self, request, pk=None):
+        """
+        Handle request to delete section through the UI;
+        deletes mentor and spacetimes along with it
+        """
+        section = get_object_or_error(self.get_queryset(), pk=pk)
```
edwardneo, line 127: `self.get_queryset()` only gets unbanned sections; just in case something went wrong and the coord is banned from the section, maybe use `Courses.objects.all()`.

csm_web/frontend/src/components/section/MentorSectionInfo.tsx
```diff
@@ -232,6 +251,58 @@ export default function MentorSectionInfo({
...
+  const [stage, setStage] = useState<DropSectionStage>(DropSectionStage.INITIAL);
+
+  const performDrop = () => {
+    sectionDropMutation.mutate(undefined, {
+      onSuccess: () => {
+        // console.log(sectionId)
+        setStage(DropSectionStage.DROPPED);
+      }
+    });
+  };
```
smartspot2: There should be a failure case handled here (ex. if there are students in the section). The query hook will also need to be modified to raise this error as well.

csm_web/scheduler/views/section.py
```diff
@@ -119,6 +119,43 @@ def create(self, request):
...
+            logger.error(
+                (
+                    "<Spacetime Deletion:Failure> Could not delete spacetime, user %s"
+                    " does not have proper permissions"
+                ),
+                log_str(request.user),
+            )
+            raise PermissionDenied(
+                "You must be a coordinator to delete this spacetime!"
+            )
```
edwardneo, line 149: Preferably the coordinator check comes before the student count check.

csm_web/scheduler/views/section.py
```diff
@@ -119,6 +119,43 @@ def create(self, request):
...
+        # Delete all spacetimes in the section
+        for spacetime in section.spacetimes.all():
+            spacetime.delete()
+        # Delete the mentor
+        section.mentor.delete()
+
+        section.delete()
```
edwardneo, line 156: This should be atomic, just in case something goes wrong.

csm_web/scheduler/views/section.py
```diff
@@ -119,6 +119,43 @@ def create(self, request):
...
+        course = section.mentor.course
+        # If the course has students, we cannot delete the section
+        if section.students.count() > 0:
+            logger.error(
+                (
+                    "<Section Deletion:Failure> Could not delete section %s, it has"
+                    " students. Remove all students manually first."
+                ),
+                log_str(section),
+            )
```
smartspot2, line 137: This should return an error with its respective error code (likely a 401 or 422). At the moment, it'll just keep going through the method.

edwardneo (changes requested):

csm_web/scheduler/tests/models/test_section.py
```diff
@@ -0,0 +1,77 @@
...
+    CoordinatorFactory.create(user=coordinator_user, course=course)
+
+    return (
+        section_one,
+        coordinator_user,
+    )
+
+
+@pytest.mark.django_db
+def test_section_delete(client, setup):
```
smartspot2, line 39: There should be additional tests for edge cases, i.e. a section that the user is not a coordinator for, if the user is not a coordinator at all, if the section has existing students, etc.

edwardneo:
Adding another test for student count blocking section deletion would be good! After that, rebase and squash this all down into one commit.

smartspot2 (changes requested):
Several small things here; most important ones are additional tests, and adding failure cases in the frontend.

There are also some linting warnings in the JS; mostly about unused variables, which should be removed in the code. There is also a lot of commented out code that should be deleted as well.

Lastly, there are still 11 commits here, which should be rebased and squashed down to one commit (or two) before merging (feel free to wait until these reviews are resolved prior to squashing).

PR #444, opened 2023-09-19
Only untaken attendances are deleted when student is dropped

Veriny:
When students are dropped, they lose their attendances for the current day, even if their attendance has already been taken. This change makes it so that attendances for the day are only deleted when the section is later in the day (i.e. attendance has not been taken)


smartspot2 (changes requested):
At the moment, this doesn't actually fix the issue, since the mentor could still have left attendances blank when the student drops the section; we should only delete attendances if they're actually in the future. This would require some additional checking for attendances on the day of the drop, to check whether the user dropped the section after the section time or not.

PR #445, opened 2023-10-06
Add stylelint, fix eslint, refactor resource aggregation styles

smartspot2:
The ESLint pre-commit hook is currently broken due to bad configuration; this PR fixes the configuration.

The CSS files are also currently not linted; stylelint was added to lint the CSS files. In the process of adding stylelint, a lot of the current CSS files fail the linter, and those files were fixed as well. This also involved refactoring resource aggregation class and ID names, since they used camel case rather than kebab case.

The current stylelint configuration also automatically sorts the CSS properties to fit a semantic ordering (currently all the CSS is ordered arbitrarily).

The pre-commit hooks have also been updated in this process of fixing the pre-commit configuration.


edwardneo (approved):
The CSS looks good, and ESLint is running!

ericyche (approved):

PR #446, opened 2023-10-09
Fix various typescript and eslint issues

smartspot2:
This PR contains a few fixes to typescript code:
* The `ROLES` object was unconventionally put in the `section/Section.tsx` file, when it should be an enum in `utils/types.tsx`
* Fixed ESLint warnings about unused variables; added specification to ignore unused variables if they're prefixed by `_`
* Added automatic import sorting for typescript files through `eslint-plugin-import`; this plugin also adds several checks to import statements
	* Fixed all import sorting in existing files. Prior, there were comments marking groups of images/styles, but since it is now automated, this is no longer needed. Convention-wise, new lines should still be put between groups, but this isn't strictly enforced).


ericyche (dismissed):

ericyche (approved):
ty for the big refactors - works on my end and doesn't break dev + commits

PR #450, opened 2023-10-17
Prevent students from taking other students' attendances

Veriny:
Before, we only checked whether the requesting user was the student that the attendance was being taken for, so savvy students could take attendance for their friends by directly sending an API request. Now, only mentors for the specific section are able to take attendance.

Fixed pylint stuff for student.py

### DIFF
--- a/csm_web/scheduler/views/student.py
+++ b/csm_web/scheduler/views/student.py
@@ -1,16 +1,14 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Q
 from django.utils import timezone
-from scheduler.models import Attendance, SectionOccurrence
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
-import datetime
 
-from .utils import log_str, logger, get_object_or_error
 from ..models import Student
 from ..serializers import AttendanceSerializer, StudentSerializer
+from .utils import get_object_or_error, logger
 
 
 class StudentViewSet(viewsets.GenericViewSet):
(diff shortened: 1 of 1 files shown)


smartspot2:
Looking at this code a little bit more, I think there isn't actually any issues with the way the request checks are handled right now; see the `get_queryset()` method of the viewset:
https://github.com/csmberkeley/csm_web/blob/0af4cb6959d75c002301bae60aa1ce38a1bc8248/csm_web/scheduler/views/student.py#L19-L27

This is referenced in the first line of the `attendances()` method when a user tries to modify the student's attendance:
https://github.com/csmberkeley/csm_web/blob/0af4cb6959d75c002301bae60aa1ce38a1bc8248/csm_web/scheduler/views/student.py#L61-L63

In particular, this queryset filters for students in any of the following categories:
* The request user's student profiles
* The students in the request user's sections, for which the request user is a mentor for
* The students in the request user's course, for which the request user is a coordinator for

It's completely valid for the request user to modify any of the students in the last two points, and the students in the first point are filtered out by the check in the method.

Further, since student objects are unique across courses (i.e. a new student object is created if a user is in a section in a different course), there also should not be any issues with any of these student objects giving the request user too much control over their attendances. That is, a situation where the request user both a mentor and a peer for a student is impossible, since these relationships would have to occur in different courses—a user cannot be a mentor and a student of the same course.

Feel free to correct me if I'm wrong though, but on the current master branch I've tried using Postman to submit new requests for student attendances and I'm met with a 403 permission denied error due to the queryset filtering.

This PR is probably fine to keep and merge in though (with the spelling mistakes fixed), to add additional redundancies in case any of the above logic changes in the future; I'd be open to discussion about this as well.

edwardneo (changes requested):
Overall looks good, just a couple spelling mistakes, and the branch name should be fixed too (spelling mistake + not in the formate `fix/*`).

edwardneo (approved):
Nice!

ericyche:
logic + spelling lgtm

PR #452, opened 2023-10-30
Fix outlined icons

smartspot2:
Some icons that are not filled were incorrectly filled in by the prior SCSS transition; this should be fixed now through the addition of a new `outline` class for icons, which sets the fill to none.


jacovkim (approved):
LGTM!

edwardneo (approved):
LGTM!

PR #456, opened 2023-10-31
fix for coord adding nonexistent user

joshualiao:
In model.py, added a check to see if the user is a coordinator in the course.

Refactored the coordinator add when the student is not in the course. It now has the following functionality:
- If the student is a non-existent user and the course is not restricted, create a new user and enroll. 
- If the student does exist, ensure they aren't associated with the course and the course is not restricted, then enroll them.


smartspot2:
What's the status on this PR? Some of the more recent changes in master should fix some of the Cypress issues I think you were having.

Additionally, there are some syntax errors when the current code is run (check the changes in `views/section.py`), which should be fixed before this is merged.

PR #457, opened 2023-10-31
Move django-heroku to prod dependencies, update requirements

smartspot2:
There are a few issues with the poetry dependency groups that I've spotted a little while ago; the main issue is that the dependency groups are not doing anything. Currently, _all_ dependencies will be installed _regardless_ of what group they are in, since none of the groups are marked as optional. This means that when installing with `poetry install` (even with `poetry install --with=dev`), all the dependencies across dev and prod are installed.

This PR marks the prod dependency group as optional, which fixes the above issue. The export command should still say the same (the `--with=prod` now actually does something). Further, since the dev dependency group remains not optional, just a simple `poetry install --no-root` would install all the main and dev dependencies; the prior `--with=dev` does nothing.

There is also currently a requirement for `psycopg` for local installation, which is technically not needed anymore with Docker; further, this is a child dependency of `django-heroku`, which is also not needed for non-prod environments. People will need to install all the poetry dependencies locally to run `pre-commit`, which means that they'll be required to install PostgreSQL for `pre-commit`, which is a large pain that we wanted to avoid through Docker.

This PR moves the `django-heroku` dependency into the prod dependencies list, which removes it from the dependencies when installing through `poetry install`.

In the process, some of the dependencies have been updated, and pre-commit hooks have been updated as well.


ericyche (approved):
ak

PR #462, opened 2023-12-23
Data Export

smartspot2:
The current data export modal is very bare-bones, and does not actually export much information.

This PR intends to overhaul the data export features, enabling coordinators to export data about courses, attendance, students, sections, etc. with a lot of control over what exactly to include in the exported CSV file.


csm_web/scheduler/views/export.py
```diff
@@ -0,0 +1,590 @@
...
+    if preview is not None and preview > 0:
+        # limit queryset
+        section_queryset = section_queryset[:preview]
+
+    # query database for values
+    values = section_queryset.values(*export_fields)
+
+    section_time_dict = {}
+    if "section_times" in fields:
+        used_ids = set(d["id"] for d in values)
```
edwardneo, line 434: Section ID bug probably has to do with this line.

csm_web/scheduler/views/export.py
```diff
@@ -0,0 +1,590 @@
...
+
+    export_type = request.query_params.get("type", None)
+    courses_str = request.query_params.get("courses", None)
+    fields_str = request.query_params.get("fields", "")
+    preview = request.query_params.get("preview", None)
+
+    if courses_str is None or export_type is None:
+        raise BadRequest(
+            "Must include `courses` and `type` fields in the query parameters"
+        )
```
edwardneo, line 40: Maybe make this a comprehendible error on the frontend if someone doesn't select any courses rather than defaulting to BadRequest? It could also not error and instead return a blank file.

edwardneo (changes requested):
Some bugs to fix:
- When section ID is unselected, no data exports
- When no course is selected, data still generates, but will error on export.

csm_web/scheduler/views/export.py
smartspot2: Reworked the frontend query to take care of this case individually.

edwardneo (approved):
Overall, looks good to me -- no bugs that break the interface anymore, but some QOL changes that would be good to add in my opinion.

edwardneo (changes requested):
Oops, I didn't mean to approve haha.

csm_web/frontend/src/components/data_export/ExportPage.tsx
```diff
@@ -0,0 +1,253 @@
...
+                <i>(More rows clipped)</i>
+              </td>
+            </tr>
+          )}
+        </tbody>
+      </table>
+    );
+  } else {
+    // not enough data
+    dataTable = <i>Query returned no data.</i>;
```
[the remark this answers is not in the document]
smartspot2: Modified to resolve that issue; it's a little bit hard to check for no returned data when >= 1 course is selected, since it's probably not the best to use preview data as a proxy for the real data (since preview data is supposed to be kinda dummy data).

So at the moment, it will download an empty CSV (with headers) if there is no data, and the button is grayed out if no courses are selected, to prevent the errors.

jacovkim (dismissed):
Looks good!

edwardneo (approved):
Aproof!

jacovkim (approved):
good!

PR #470, opened 2024-03-04
Restyle matcher UI

### LINKED ISSUE #437: Matcher UI should indicate linked sections more prominently
It's currently very easy to accidentally have sections unintentionally be linked together, causing a lot of problems down the road when some mentors have put preferences for the linked sections. At that point, the only way to solve the issue is to delete all slots and preferences and start over.

To prevent this from occurring in the first place, we can make it a lot more prominent when slots are linked together (perhaps with a tooltip, or with some other visible text), or we can have a list of all of the sections in text form somewhere prior to opening the preference form to the mentors.

Both could be good to add, but in general it should be made clearer what exactly the slots are before allowing mentors to add preferences, since once the form is open, it becomes very hard to change the slots.

smartspot2:
This PR makes various miscellaneous improvements to the matcher UI.

Coordinators have created slots without knowing that many slots are actually linked; a confirmation modal has been added because of this, listing out all of the slots along with a count of the number of slots to submit.

Another confirmation modal was added to the button opening the preference form for mentors, making it clear that slots cannot be edited afterward. This can probably be reverted or changed if in the future we decide to rework the slot creation system to allow for slot edits.

The create stage button UI/UX was also modified; to avoid any confusion with save/submit/continue buttons, the flows were reworked to have only one green (primary) button visible at any given point in time. Clicking this button would proceed with the flow, whereas secondary buttons either cancel the flow or go to a different flow (ex. editing events).

Further, there were many matcher button and input styles that are not consistent with the rest of the site; these classes have also been migrated over.

Closes #437.

### DIFF
--- a/csm_web/frontend/src/components/enrollment_automation/MentorSectionPreferences.tsx
+++ b/csm_web/frontend/src/components/enrollment_automation/MentorSectionPreferences.tsx
@@ -188,10 +188,10 @@ export function MentorSectionPreferences({
             // multiple selected events
             <div className="matcher-sidebar-header">Multiple selected events</div>
           )}
-          <label>
+          <label className="matcher-pref-input-label">
             Preference:
             <input
-              className="matcher-input"
+              className="form-input light"
               type="number"
               key={event.id}
               defaultValue={event.preference}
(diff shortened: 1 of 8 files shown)


jacovkim (approved):
Looks good to me!

edwardneo (changes requested):
Looks generally good -- why is there a save button and a submit button for creating the matcher slots, and you can hit submit without saving? It seems a bit weird to me, and I think coordinators will accidentally forget to save a slot they wanted to save but submit it anyways.

edwardneo (approved):
LGTM!

jacovkim (approved):
LGTM!

PR #476, opened 2024-07-13
Update docker-compose to allow migrations to be created in Docker

smartspot2:
Currently, running `docker compose exec django makemigrations` will error, since the Django file system in Docker is read-only. This PR modifies the bind mounts to allow writes to the migrations folder. (The migrations folder in the frontend app is not included here, since we only make database changes within the `scheduler` app.)


edwardneo (approved):
LGTM!

PR #477, opened 2024-07-13
Upgrade to Python 3.12 and Django 5

### LINKED ISSUE #453: Upgrade Python version
The current Python version used in this repository is quite outdated; Python 3.9.13 was released in 2020, and 3.9.13 (released May 2022) was the last maintenance update for Python 3.9. The current latest 3.9 version is 3.9.18, with a bunch of security updates in between.

We should probably update the Python version to 3.11 or 3.12 to ensure that we continue getting the newest features and support from the latest packages. This upgrade may cause some dependency issues if any of our required dependencies have not added support for the later Python versions, though most should be updated to 3.11 now (3.12 may be more of an ask, since it's still quite new).

This is also a pretty good chance to update as many Python dependencies as we can, especially for security updates and other bugfix updates in any packages.

smartspot2:
Closes #453

With the release of Django 5.0, Python 3.9 is no longer supported. Python 3.9 is also incredibly old at this point, so updating to a more recent version can be beneficial for the extra features and fixes.

This PR upgrades dependencies to support Django 5 and Python 3.12.

The most notable changes are:
* All GitHub CI, Docker containers, and Heroku files have been updated to use Python 3.12 instead of 3.9.
* Django 4.2 deprecated the `DEFAULT_FILE_STORAGE` and `STATICFILES_STORAGE` options, adding a new `STORAGES` option instead. The existing settings have been updated to use this new setting.
* All existing dependencies have been updated to the most recent versions listed on PyPI.
* Django 4.1 deprecated logging out with a `GET` request, and the ability to do so is removed in Django 5.0; this means that several places need to be adjusted to use `POST` requests instead. In particular:
    * The logout button now calls a utility method to send a `POST` request (following redirects)
    * A new Cypress command for logging out a user has been added, and existing Cypress tests that have a step logging users out were also updated. (New tests have been added to ensure that no regressions occur in the future.)


edwardneo (approved):
Tested with current scheduler data populated and works fine. Make sure to check docker containers on Heroku are updated properly. LGTM!

PR #478, opened 2024-07-26
Replace django get_random_secret_key with secrets.token_hex

smartspot2:
With prior dependency changes, we no longer install `psycopg` onto users' computers, but the `get_random_secret_key` function used in `setup.sh` requires Django to be fully set up prior to calling the function. As such, the function errors due to the missing dependency.

Since the secret key generated here is only going to be used for development, we can simply replace it with a more straightforward `secrets.token_hex(32)`, despite its much lesser security.


smartspot2:
Merged the PR without review since it's just a development setup change :)

PR #480, opened 2024-08-12
Add slot descriptions, matcher randomness

### LINKED ISSUE #466: Add randomness to matcher algorithm
Currently, the matcher tends to output the same tiebreak, but we would prefer some variety between courses and section times for easier room bookings. There are some solutions to this (consider matcher runs from courses already done, randomize the output without seeding), but the likely best solution is to randomize tiebreaks using the course ID as a seed to consistently output the same schedule for the same course but a different schedule for a different course, even if the courses have the same preferences. Although this doesn't guarantee the least scheduling overlap, in expectation this should help room booking.

smartspot2:
Closes #466.

Matcher slots currently have no special metadata that coordinators can use to indicate special sections with---this information needs to be communicated to mentors separately. Adding a description to slots allows for this metadata to be embedded within the matcher, so mentors can see whether sections have any special tags (ex. NPE, affinity section, etc.) prior to filling availability.

Secondly, this PR also adds some randomness to the matcher algorithm, randomly shuffling the lists of mentors/slots/preferences so that tiebreaks happen approximately uniformly.


edwardneo (changes requested):
Some minor things. I'm not sure which should be addressed in this PR or another PR.
- The precommit check is failing.
- It would be nice if the mentor side of the matcher had the descriptions on the tiles instead of only visible if you click on the tile -- otherwise, mentors have to click through the tiles to find the description. Or some other visual indicator of a description.
- The save button on the section edit page is always gray -- I think it should be green.
- The bulk editing feature on the section edit page only edits capacity and description, not section time. Should it edit section time?

isabellaalps (changes requested):
note: this PR should be prioritized over calendar as we need the description changes to be merged in before elaine and i make too much progress on the affinity section pop-up

+1 to edward's comments

smartspot2:
Newest changes should resolve the aforementioned comments:
* Description is now displayed on the calendar event
	* I currently have it implemented where the description is displayed in parentheses on the same line as the event time; this is so that in _most_ cases, the description will be visible, even if the event is 30 minutes long. If the schedule time or description text becomes too long, it'll overflow and get cut off, but that's just something we have to deal with regardless.
	* In the mentor view, this does mean that the preference number is more likely to get pushed out of the bounding box of the event, but again we'd have this issue regardless of the description if the event is 30 minutes long.
	* Not sure if there are any opinions on this; a more robust solution would be to have a larger bounding box on hover, though this would take more work, and I'm not sure if it's completely worth it at this point.
* Save button is now styled with `primary-btn`
* Bulk editing has now been fixed to also work with section times (this was probably missed in the initial implementation)

Regarding the `pre-commit` check, this is a more fundamental issue with the pylint configuration; it seems very weird that it's only failing for the fields in the `DateTimeField` models (and other similar time-related models), but after an update to `pylint-django`, it seems to have gone away? I've created #521 for that dependency update.

For now, we can just ignore the pre-commit failures in this PR, and #512 can be merged into master first if we want to maintain the passing CI on master.

isabellaalps (approved):
looks good to me :) ty Alec

ElaineShu0312 (approved):
excellent work alec

ElaineShu0312 (approved):
excellent work alec

edwardneo (approved):
Ok LGTM! I'll put more QOL feedback on another issue for someone to one day implement...

PR #501, opened 2024-10-21
Fix queries for counts in course data export

smartspot2:
Prior, due to traversals through many-to-many and many-to-one field relations, the counts for the number of students/mentors/sections were messed up. This has now been fixed by ensuring that the counts are for distinct objects.

The Django admin console section count has also been fixed to actually look at the number of sections, since now mentors can be associated with no section.


jacovkim (approved):
LGTM!

PR #502, opened 2024-10-28
refactors and encapsulates some logic, updates model func but incomplete

faithdennis:
merges Gabe's work with mine, incorporates some feedback from Edward, removes some model logic out of views and encapsulates model logic. needs some changes based on design direction.

### DIFF
--- a/csm_web/scheduler/models.py
+++ b/csm_web/scheduler/models.py
@@ -15,6 +15,7 @@ logger = logging.getLogger(__name__)
 
 logger.info = logger.warning
 
+DEFAULT_WAITLIST_CAP = 3
 
 class DayOfWeekField(models.Field):
     DAYS = (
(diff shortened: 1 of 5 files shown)


csm_web/scheduler/views/waitlistedStudent.py
```diff
@@ -17,91 +17,69 @@ def add(request, pk=None):
     """
     Endpoint: /api/waitlistedstudent/add
```
smartspot2, line 18: The docstrings here haven't been updated to reflect the URL changes from `urls.py`.

csm_web/scheduler/models.py
```diff
@@ -232,19 +254,36 @@ class WaitlistedStudent(Profile):
...
+                    logger.info("Student added from waitlist successfully")
+                
+                waitlisted_student.active = False  # Mark the waitlisted student as inactive similar to drop student
+                student.save()
+                # TODO: update positions of all other waitlisted students
+
+            # TODO: confusion around whether we set active = False or delete
+            student_waitlist = course_queryset.filter(user=student.user)
+            for waitlist in student_waitlist:
+                waitlist.delete()
```
smartspot2, line 286: There are some TODOs here that should probably be resolved at some point (no need to do that here if that's not the plan).

Things to think about:
* This can probably be extracted to its own helper function as a pure python function, since it'll be called by multiple places.
* You can specify an ordering within the model itself (with the `ordering` field in an inner `Meta` class), so that you do not need to worry about implementing a sort here.
* The ordering here would be by timestamp, which would work fine, unless we want to allow coordinators to re-order the waitlist. This isn't something to worry about now, but an alternative is to just associate an integer with each student, and preserve the invariant that sorting by these integers gives the order in the waitlist.

PR #506, opened 2024-11-01
Initial Waitlisting Feature Dev 

gabrielhan23:
Finished initial backend feature for waitlisting. 
- waitlist model (using timestamps)
- add and drop api for waitlist
- auto add from waitlist after students drop
- light testing with api and admin tools

Not implemented yet:
- coord capabilities
- what to do when student being added from waitlist fails
- waitlist ordering


anonymized_data.json
smartspot2, line 1: This file probably shouldn't be committed

docker-compose.yml
```diff
@@ -33,8 +36,8 @@ services:
         source: ./package-lock.json
         target: /opt/csm_web/package-lock.json
         read_only: true
-      # prevent node modules from being overwritten on host
-      - notused:/opt/csm_web/app/node_modules
+    command: sh -c "npm install && npm start"
+
```
smartspot2, line 40: What's this change for? The entry point should already have done all the setup and server initialization.

(If you revert this, make sure all the changes in this file are reverted.)

smartspot2 (changes requested):
Haven't taken too close of a look at the implementation yet, but just wanted to point out some things I noticed; left some comments about them.

In addition, try to at least get the pytests and the pre-commit check to pass (I don't believe that they'll automatically run on this PR since it's not targeting master, so you may need to manually re-run tests on the PR).

csm_web/scheduler/tests/models/test_waitlisted_student.py
faithdennis, line 1: why are we leaving these comments in?

faithdennis (approved):

faithdennis (commented):
Some tests seem to be failing according to the automated tests. Could changing section view have broken other tests? If not, I'm not sure if these automated tests are valid.

smartspot2:
> Some tests seem to be failing according to the automated tests. Could changing section view have broken other tests? If not, I'm not sure if these automated tests are valid.

pre-commit is failing due to the `anonymized_data.json` file being incorrectly committed, and pytest/cypress tests are failing due to the following error:
```
Traceback (most recent call last):

  File "/opt/hostedtoolcache/Python/3.9.20/x64/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute

return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: null value in column "max_waitlist" of relation "scheduler_course" violates not-null constraint

DETAIL:  Failing row contains (1, CS61A, 2024-12-01, 2024-10-02 16:03:44.763658+00, 2024-11-16 17:03:44.763658+00, 2, Structure and Interpretation of Computer Programs, 2024-10-17, f, null, null).
```
Seems like you need to make sure that `max_waitlist` is set for all `Course` models in tests, since it's set as non-null.

anonymized_data.json
gabrielhan23: removed file from commit!

docker-compose.yml
gabrielhan23: i think it was changed in the very first base commit fixed now!

PR #513, opened 2024-12-05
Image validation and size optimization for Profile image upload from both front-end and back-end

jacovkim:
Added validation for extension and file size before reading.
Added validation after reading.
Added file size compression.

Some other considerations:
1. Crop the image after resizing to preserve the ratio
2. Placeholder image for missing profiles
3. Might need to replace the current compression method as it's an iterative approach
4. (Optional) Change public S3 bucket to private with AWS signing

Things to do:
1. Should disable the "upload" button when the file exceeds the limit when editing
2. Also clear form after uploading


smartspot2 (changes requested):
I just did a cursory skim over the PR, but here's some of the feedback I have so far.

Also, did you mean for the target branch to be `master` here, or is this supposed to be targeting a main feature branch?

csm_web/scheduler/views/user.py
```diff
@@ -23,12 +32,239 @@ def list(self, request):
...
+        buffer.seek(0)
+        if img_format == "JPEG" or img_format == "JPG":
+            image.save(buffer, format=img_format, quality=quality)
+        else:
+            image.save(buffer, format="PNG", optimze=True)
+        if buffer.tell() <= target_size_bytes or quality <= 50:
+            break
+        quality -= 5  # decrease quality to reduce file size
+    buffer.seek(0)
+    return buffer
```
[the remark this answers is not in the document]
jacovkim: usually, selfies or phone images tend to be larger than 5MB so allowing 10MB input size and compressing it to 5MB was what I was initially thinking but just realized I didn't change the target size and input limit. But should allowing 10MB then compressing to 5MB to save space make sense?

csm_web/frontend/src/components/UserProfile.tsx
```diff
@@ -0,0 +1,323 @@
...
+                type="text"
+                id="pronunciation"
+                name="pronunciation"
+                maxLength={50}
+                className="form-input user-profile-input-field"
+                value={formData.pronunciation}
+                onChange={handleInputChange}
+              />
+            </div>
+            <div className="user-profile-input">
```
isabellaalps, line 277: you can't actually edit this field, so having it appear (especially appear without the email address auto-filled in) can be confusing
-note: more of a UI/UX issue than a functionality issue

csm_web/frontend/src/components/UserProfile.tsx
```diff
@@ -0,0 +1,323 @@
...
+              <label htmlFor="email" className="form-label">
+                Email
+              </label>
+              <input
+                type="text"
+                id="email"
+                name="email"
+                className="form-input user-profile-input-field"
+                value={requestedData?.email}
+                disabled
```
isabellaalps, line 287: you can't actually edit this field, so having it appear (especially appear without the email address auto-filled in) can be confusing
-note: more of a UI/UX issue than a functionality issue

isabellaalps: ignore, think it was my fault that the email didn't appear? did not create test data before running server and the email field was empty. email address was added after running createtestdata command

isabellaalps:
- think it would be nice to add a short description/heading on the profiles landing page to explain what profiles is/does
- unable to add image to profiles without triggering fatal error

csm_web/frontend/src/components/course/SectionCard.tsx
```diff
@@ -90,13 +90,17 @@ export const SectionCard = ({
...
           <CheckCircle height={iconHeight} width={iconWidth} />
           <h3>Successfully enrolled</h3>
-          <ModalCloser>
-            <button className="primary-btn">OK</button>
-          </ModalCloser>
+          <h4>To view and update your profile, click the button below</h4>
+          <Link className="primary-btn" to="/profile">
+            <UserIcon width={inlineIconWidth} height={inlineIconHeight} />
+            Profile
+          </Link>
```
smartspot2, line 103: This change is the cause of the Cypress test failures; the Cypress tests should change to look for `Profile` instead, and should make sure that the user is redirected as desired.

smartspot2 (changes requested):
Made a bunch of comments in the thread. Overall, looks pretty good!

Some other miscellaneous things as I tested manually:
* Coordinators are able to edit anybody else's profile, which shouldn't be happening; I mentioned this as a review comment as well pointing to the section of code responsible. In general, there should be additional testing to ensure that negative constraints are satisfied. (To test, I had to remove the superuser exception in the code.)
* Putting a blank name does not default to the user's full name. Interestingly, this only occurs when there is no image uploaded. Otherwise, the name defaults as usual. (This is probably because of some query invalidation to display the image once it's uploaded.)
* I think there should be a button to clear the image if the user wants to delete their profile picture; currently, there is no way to do that, and if a user wants to change their profile picture to the default, they'd need to upload the CSM logo. For initial release, this can probably be deprioritized, but this can probably be relatively easily implemented by adding an "X" icon overlaid with the uplaod icon, and the PUT request can specify `null` for the file (the backend would then check if the file is `null`; a file that is not updated would not have a key in the request).
* The upload icon is a little bit faint when there's an image uploaded; I think either (a) make the icon less transparent, or (b) put a white background on the icon so that the image itself is a little more faint, bringing out the icon more.
	<details><summary>(for example)</summary>
	
	image
	
	</details>
* The links look a little bit out of place; styling the links to fit in a little more with the color scheme could be good? The links on the resources page for worksheets could be good to model off of here. If you'd like, you can use a lighter blue (kinda like the csmentors.org website links) instead of the green to make it more clearly link-like. One big thing is that there shouldn't be a difference between normal links and `:visited` links on the page.
    <details><summary>(for example)</summary>

    image
    image

    </details>

edwardneo:
Ok I addressed most of y'all's comments. There is a working delete button, hyperlink colors are fixed, editing should be fixed. Theoretically this feature is fully functional, but if/when I have time tonight, I'll add a tooltip on the profile landing page to explain what profiles does (like Isabella mentioned) and maybe try adding a page to list all of the mentors for a certain course.

Also, the S3 bucket is currently `csm-web-profile-pictures-dev`, this can be set in the `.env` file with the link
```AWS_PROFILE_PICTURE_BUCKET_NAME=csm-web-profile-pictures-dev```
or it should have been also added to `./setup.sh` if y'all would prefer to run that. On prod it will be set manually to the prod profile pictures bucket.

File: csm_web/frontend/src/components/course/CourseBios.tsx [hunk 5]
```diff
@@ -0,0 +1,44 @@
...
+
+interface UserWithSectionTimes {
+  user: RawUserInfo;
+  sectionTimes: Spacetime[][];
+}
+
+const CourseBios = ({ sectionsByDay }: CourseBiosProps) => {
+  const flattenedSections = sectionsByDay.reduce((acc, curr) => acc.concat(curr.sections), [] as Section[]);
+  const users: Record<number, UserWithSectionTimes> = {};
+  const userOrder: number[] = [];
```
smartspot2, line 19: Since this uses the users provided from the course sections query, a user updating their profile will not invalidate the underlying query, and the bios page will not be updated until the page is refreshed.

Ex. if you update someone's bio, then navigate to the bios page without refreshing, the update will not show.

csm_web/frontend/src/components/course/BioEntry.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+
+        <div className="profile-item">
+          <p className="profile-text profile-email">{user.email}</p>
+        </div>
+
+        <div className="profile-section-times">
+          {sectionTimes.map((spacetimes, index) => (
+            <div key={index} className="profile-section-time">
+              <ClockIcon /> <span>{spacetimes.map(formatSpacetimeInterval).join(" + ")}</span>
```
smartspot2, line 53: Having `+` be the separator looks weird; it's better to just make each spacetime on its own line, similar to the section cards. (Make sure the clock icon is vertically centered though.)

csm_web/frontend/src/css/profile.scss
```diff
@@ -0,0 +1,224 @@
...
+  color: $csm-neutral-darkened;
+}
+
+.user-profile-bio-container {
+  display: flex;
+  flex-direction: column;
+  gap: 0.2rem;
+  align-items: flex-start;
+  width: 100%;
+}
```
smartspot2, line 210: There could probably be a little more space before the bio container; visually seems a little too close to the email input

image

csm_web/frontend/src/components/course/BioEntry.tsx
smartspot2, line 53: Also I think it'd be pretty cool if clicking on the section time (or a link adjacent) would bring you directly to the section card on the sections page, for ease of enrollment.

Or honestly some way of combining the bios page and sections page in the future would probably good too, ex. incorporating the enroll button on the bios page for reduced friction.

These are all nice-to-have things as extra additions for the future though (feel free to separate this into a separate issue to tackle later).

csm_web/frontend/src/components/UserProfile.tsx
smartspot2, line 1: Something for the future, but some kind of back button to return the user to their previous page could be nice here (there should be some utilities to just pop from the navigation history); having to navigate back using the browser back button is a little bit annoying. After looking at a profile, I naturally look for a back button to go back to what I was looking at before, ex. the sections page, or the bios page.

Not a priority, so this can also be punted to a separate issue.

csm_web/scheduler/urls.py
```diff
@@ -15,7 +15,10 @@
 urlpatterns = router.urls
 
 urlpatterns += [
-    path("userinfo/", views.userinfo, name="userinfo"),
+    path("user/", views.user_info, name="user"),
+    path("user/<int:pk>/", views.user_retrieve, name="user_retrieve"),
+    path("user/<int:pk>/update/", views.user_update, name="user_update"),
+    # path("user/upload_image/", views.upload_image, name="upload_image"),
```
smartspot2, line 21: Comment should be deleted here.

csm_web/scheduler/views/__init__.py
```diff
@@ -6,4 +6,6 @@
 from .section import SectionViewSet
 from .spacetime import SpacetimeViewSet
 from .student import StudentViewSet
-from .user import UserViewSet, userinfo
+
+# from .test import upload_image
```
smartspot2, line 10: Comment should be deleted here.

smartspot2 (dismissed):
Honestly overall this looks pretty good. I'll give a stamp of approval for now, but see my comments for various things I've noticed while testing.

Make sure that all conflicts are resolved and all CI tests pass, and additionally make sure that this is tested on staging before merging into master; check all permissions and do spot checks on functionality to ensure that nothing weird happens on staging. (It's also a good way to check that dependencies are all fine, since you've updated some dependencies in this PR.)

Issue #518, opened 2025-02-11
Remove usage of django-heroku


smartspot2:
The `django-heroku` package is deprecated, as of October 15, 2024. The package doesn't really do too much, other than set a few configuration parameters, which we can definitely do ourselves.

See https://github.com/heroku/python-getting-started/blob/main/gettingstarted/settings.py for an example of a Django `settings.py` file for use in Heroku. Further, see https://github.com/heroku/django-heroku/blob/master/django_heroku/core.py for the configuration that `django_heroku` adds, which we should inline in our `settings.py`.

PR #519, opened 2025-02-11
Disable configuration of staticfiles for heroku

smartspot2:
The prior integration of Python 3.12 and Django 5.x broke the build process on Heroku; this PR fixes the issues.

In the future, we should look into removing `django-heroku` entirely, since the package is deprecated. (See #518.)


smartspot2:
Merging this in myself after CI passes, since it's a small change, and has already been tested on `csm-web-staging`.

PR #521, opened 2025-02-25
Update python dependencies, fix new pylint warning

smartspot2:
Routine upgrade of python dependencies from Poetry; this PR also fixes a pylint warning in `models.py` (likely popped up with an upgrade of pylint or pylint-django).


ElaineShu0312 (approved):
**excellent** work alec

edwardneo (approved):
LGTM!

